PR #9, opened 2014-09-18
Feature/account product fiscal classification migration 8.0

renatonlima:


rvalyi:
we will analyse the work of Sylvain Legal before freezing this module on v8.

__unported__/account_fiscal_position_rule_purchase/security/account_fiscal_position_rule_purchase_security.xml
```diff
@@ -1,8 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
 
-    <data>
+	<data noupdate="1">
 
-    </data>
+		<record id="account_move_comp_rule" model="ir.rule">
+			<field name="name">Account Product Fiscal Classification</field>
+			<field name="model_id" ref="model_account_product_fiscal_classification"/>
```
danimaribeiro, line 8: This model exists in module account_product_fiscal_classification, and since this module is not dependent of that, this should be placed somewhere else.

rvalyi:
this is nuts, it seems we have the same errors in the test as we have on HEAD with the account_fiscal_position_rule module (not involved in this merge). It's strange because locally I got this test passing and I really believe it was green also when I did the merge of this module. I'll investigate this.

gurneyalex:
CLA ok

rvalyi:
Hello, as for v8, as discussed with @legalsylvain , we may rather merge his product_taxes_group instead of account_fiscal_classification. This is unfortunate as it was a double work from @legalsylvain who didn't know the existence of this module long ago. But his module is really better coded than ours that we did a bit at the early stages.
Here is the other module we would ask to switch project then:
https://github.com/OCA/account-financial-tools/pull/32

I think this module a dependency only of the Brazilian localization at the account localization level, so it would need some refactor, but I think we should take advantage of v8 migration to have an upstream all clean in the BR localization. I'll evaluate a bit the require refactor work and eventually ask or contrib a few change to @legalsylvain 's module.

legalsylvain:
Hi, 
Thanks @rvalyi. I'll do the PR in V8 the next week.
Again this PR, please say me what I can change to make more easy the migration.

Kind regards.

danimaribeiro:
This could be closed.

PR #10, opened 2014-09-19
Feature/account fiscal position rule migration 8.0

renatonlima:


account_fiscal_position_rule/account_fiscal_position_rule.py
```diff
@@ -0,0 +1,260 @@
...
+            addrs['invoice'] = self.env['res.partner'].browse(
+                partner_invoice_id)
+
+        # In picking case the invoice_id can be empty but we need a
+        # value I only see this case, maybe we can move this code in
+        # fiscal_stock_rule
+        else:
+            partner_addr = partner.address_get(['invoice'])
+            addr_id = partner_addr['invoice'] and \
+                partner_addr['invoice'] or None
```
guewen: Please do not use `... and ... or ...` but `... if ... else ...`, and do not use `\` but parenthesis.

account_fiscal_position_rule/account_fiscal_position_rule.py
```diff
@@ -0,0 +1,260 @@
...
+        result['value'].update(self.fiscal_position_map(**kwargs))
+        return result
+
+
+class AccountFiscalPositionRuleTemplate(models.Model):
+    _name = 'account.fiscal.position.rule.template'
+    _description = 'Account Fiscal Position Rule Template'
+    _order = 'sequence'
+
+    name = fields.Char('Name', size=64, required=True)
```
guewen: `size` to remove: it should be set only when there is a reason to limit (such as an ean13 for instance)

account_fiscal_position_rule/account_fiscal_position_rule.py
```diff
@@ -0,0 +1,260 @@
...
+    description = fields.Char('Description', size=128)
```
guewen: `size` to remove: it should be set only when there is a reason to limit (such as an ean13 for instance)

account_fiscal_position_rule/account_fiscal_position_rule.py
```diff
@@ -0,0 +1,260 @@
...
+
+from openerp import models, fields, api
+
+
+class AccountFiscalPositionRule(models.Model):
+    _name = 'account.fiscal.position.rule'
+    _description = 'Account Fiscal Position Rule Template'
+    _order = 'sequence'
+
+    name = fields.Char('Name', size=64, required=True)
```
guewen: `size` to remove: it should be set only when there is a reason to limit (such as an ean13 for instance)

account_fiscal_position_rule/account_fiscal_position_rule.py
```diff
@@ -0,0 +1,260 @@
...
+    description = fields.Char('Description', size=128)
```
guewen: `size` to remove: it should be set only when there is a reason to limit (such as an ean13 for instance)

account_fiscal_position_rule/account_fiscal_position_rule_view.xml
```diff
@@ -0,0 +1,223 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+
+	<data>
```
guewen: Needs to be reindented in order to use 2 spaces instead of tabs.

guewen:
This module deserves tests.

guewen:
Thanks for the work on the migration

account_fiscal_position_rule/account_invoice.py
```diff
@@ -20,45 +20,44 @@
...
-class account_invoice(osv.Model):
+class AccountInvoice(models.Model):
     _inherit = 'account.invoice'
 
-    def _fiscal_position_map(self, cr, uid, result, context=None, **kwargs):
+    def _fiscal_position_map(self, result, **kwargs):
+        ctx = dict(self._context)
+        ctx.update({'use_domain': ('use_invoice', '=', True)})
+        return self.env['account.fiscal.position.rule'].with_context(
+            ctx).apply_fiscal_mapping(result, **kwargs)
```
guewen, line 33: You don't need to update the context yourself. `with_context()` does that. You can write (with a better care of the length of the line, probably by extracting a part in a variable):

```
return self.env['account.fiscal.position.rule'].with_context(use_domain=('use_invoice', '=', True)).apply_fiscal_mapping(result, **kwargs)
```

rvalyi:
@guewen yes we will put some tests before merging, we just didn't have time so far.

max3903:
:+1:

jgrandguillaume:
:+1:

PR #12, opened 2014-10-08
[ADD] Allow a fiscal position to be set on a shop.

mistotebe:
Sometimes a company is not enough to distinguish what taxes should apply, but a shop is (e.g. Magento).

### DIFF
--- a/account_fiscal_position_rule_sale_shop/__init__.py
+++ b/account_fiscal_position_rule_sale_shop/__init__.py
@@ -1,0 +1,22 @@
+# -*- encoding: utf-8 -*-
+###############################################################################
+#
+#   account_fiscal_position_rule_sale_shop for OpenERP
+#   Copyright (C) 2013-2014 credativ ltd. <http://www.credativ.co.uk>
+#     @author Ondrej Kuznik <ondrej.kuznik@credativ.co.uk>
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU Affero General Public License as
+#   published by the Free Software Foundation, either version 3 of the
+#   License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+from . import account_fiscal_position_rule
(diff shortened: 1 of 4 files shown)


account_fiscal_position_rule_sale_shop/account_fiscal_position_rule.py
```diff
@@ -0,0 +1,42 @@
...
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+from osv import fields, osv
+
+
+class account_fiscal_position_rule(osv.Model):
```
tafaRU: osv.Model is deprecated, use instead orm.Model

account_fiscal_position_rule_sale_shop/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU Affero General Public License for more details.
+#
+#   You should have received a copy of the GNU Affero General Public License
+#   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+###############################################################################
+
+import account_fiscal_position_rule
```
tafaRU: Use explicit relative import rather than absolute:

`from . import account_fiscal_position_rule`

account_fiscal_position_rule_sale_shop/security/account_fiscal_position_rule_sale_shop_security.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
pedrobaeza: Why are you including these files if they are empty?

mistotebe: Most other modules have the same structure - the two files are present even if they are empty, seemed like a convention. Should they be removed?

mistotebe:
Sorry for the old-style code, this was ported from a 6.1 module we are deconstructing.

account_fiscal_position_rule_sale_shop/security/account_fiscal_position_rule_sale_shop_security.xml
pedrobaeza: Yeah, if there is nothing to add, don't add it :wink:

rvalyi:
Hum removing the empty files that doesn't look bad. Notice however that in v8 we won't have sale.shop object anymore and you would need to come up with a whole different module (possibly using sale teams or warehouses). Now once the module is clean, why not accepting it as for v7.

mistotebe:
Any comments on that last change? It should address all comments above.

mistotebe:
Any comments?

guewen:
:+1:

craiggowing:
Any thoughts on merging?

mistotebe:
ping

pedrobaeza:
As this one was already reviewed and approved, I finally merge it.

PR #15, opened 2015-01-04
initial german translation and add missing pot (template files)

rschnapka:


account_fiscal_position_rule_stock/i18n/de.po
```diff
@@ -0,0 +1,146 @@
...
+#: field:account.fiscal.position.rule,from_state:0
+#: field:account.fiscal.position.rule,to_state:0
+#: field:account.fiscal.position.rule.template,to_state:0
+msgid "State To"
+msgstr "Status bis"
+
+#. module: account_fiscal_position_rule
+#: model:ir.module.module,description:account_fiscal_position_rule.module_meta_information
+msgid "Include a rule to decide the correct fiscal position"
+msgstr ""
```
pedrobaeza, line 71: Extra line

yvaucher:
@pedrobaeza I think this po file is valid that way

:+1:

pedrobaeza:
Merging although it includes POT and now this is handled by Transifex, but it deserves as is 7.0.

PR #17, opened 2015-03-24
account fiscal position rule stock migration 8.0

mileo:


rvalyi:
@mileo could you rebase upon 8.0 so we ensure tests are ok as a 1st check?
I didn't check yet this branch against https://github.com/OCA/account-fiscal-rule/pull/30
But I would say f we decide to take this branch we should at least cherry pick the test added by @danimaribeiro

rvalyi:
note that however as a naive approach I would prefer this branch as it's based on the one we are using in production for more than 9 months now at several customers.

max3903:
@renatonlima or @mileo Can you fix conflicts and Travis errors ? Thanks!

danimaribeiro:
#30

max3903:
@danimaribeiro can you make a pull request against the branch of this pull request ? Having 3 pull requests for the same module make it longer to review and merge. Thanks for your collaboration!

danimaribeiro:
@max3903 #30 is a continuation from another PR #27, not this branch, I created new one because bwrsandman was no longer working with Odoo.
I decided to go with that one because there were people reviewing that. This one was here without review since march.

max3903:
@danimaribeiro I just closed #27.

@rvalyi If you prefer this branch, can you help having it fixed so we can merge it first and then #30? Otherwise, we close it and review #30.

rvalyi:
@max3903 yes that's the idea, will look at both. Meanwhile I hope you see the picture of the 8.0 migration of this repo is improving quickly overall.

mileo:
What we can do about this module? I can fix the issues, but I'm not able to work for nothing.

mileo:
Please use #30

PR #20, opened 2015-05-28
[UPD] adapt code for in & out pickings, fix #19

### LINKED ISSUE #19: [7.0] account_fiscal_position_rule_stock
Steps:
- install account_fiscal_position_rule_stock
- Create an incoming shipment or a delivery order
- Select a partner, the next traceback occurs:

Server Traceback (most recent call last):
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/addons/web/session.py", line 89, in send
    return openerp.netsvc.dispatch_rpc(service_name, method, args)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/netsvc.py", line 300, in dispatch_rpc
    result = ExportService.getService(service_name).dispatch(method, params)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/service/web_services.py", line 632, in dispatch
    res = fn(db, uid, _params)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/osv/osv.py", line 193, in execute_kw
    return self.execute(db, uid, obj, method, *args, *_kw or {})
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/osv/osv.py", line 133, in wrapper
    return f(self, dbname, _args, *_kwargs)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/osv/osv.py", line 202, in execute
    res = self.execute_cr(cr, uid, obj, method, _args, *_kw)
  File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/2946724-7-0-95feb0/openerp/osv/osv.py", line 190, in execute_cr
    return getattr(object, method)(cr, uid, _args, *_kw)
TypeError: onchange_partner_in() takes at most 6 arguments (7 given)

rdeheele:
It adapts code logic working on stock.picking object (internal moves)
for out/in pickings.

### DIFF
--- a/account_fiscal_position_rule_stock/stock.py
+++ b/account_fiscal_position_rule_stock/stock.py
@@ -75,3 +75,109 @@ class stock_picking(osv.Model):
         result['fiscal_position'] = picking.fiscal_position and \
             picking.fiscal_position.id
         return result
+
+
+class stock_picking_out(osv.Model):
+    _inherit = "stock.picking.out"
+    _columns = {
+        'fiscal_position': fields.many2one(
+            'account.fiscal.position', 'Fiscal Position',
+            domain="[('fiscal_operation_id','=',fiscal_operation_id)]"),
+    }
+
+    def _fiscal_position_map(self, cr, uid, result, **kwargs):
+
+        if not kwargs.get('context', False):
+            kwargs['context'] = {}
+
+        kwargs['context'].update({'use_domain': ('use_picking', '=', True)})
+        fp_rule_obj = self.pool.get('account.fiscal.position.rule')
+        return fp_rule_obj.apply_fiscal_mapping(cr, uid, result, **kwargs)
+
+    def onchange_partner_in(self, cr, uid, ids, partner_id=None,
+                            company_id=None, context=None, **kwargs):
+
+        result = super(stock_picking_out, self).onchange_partner_in(
+            cr, uid, partner_id, context)
+
+        if not result:
+            result = {'value': {'fiscal_position': False}}
+
+        if not partner_id or not company_id:
+            return result
+
+        partner_invoice_id = self.pool.get('res.partner').address_get(
+            cr, uid, [partner_id], ['invoice'])['invoice']
+        partner_shipping_id = self.pool.get('res.partner').address_get(
+            cr, uid, [partner_id], ['delivery'])['delivery']
+
+        kwargs.update({
+            'partner_id': partner_id,
+            'partner_invoice_id': partner_invoice_id,
+            'partner_shipping_id': partner_shipping_id,
+            'company_id': company_id,
+            'context': context,
+        })
+        return self._fiscal_position_map(cr, uid, result, **kwargs)
+
+    def _prepare_invoice(self, cr, uid, picking, partner, inv_type,
+                         journal_id, context=None):
+        result = super(stock_picking_out, self).\
+            _prepare_invoice(cr, uid, picking, partner,
+                             inv_type, journal_id, context)
+        result['fiscal_position'] = picking.fiscal_position and \
+            picking.fiscal_position.id
+        return result
+
+
+class stock_picking_in(osv.Model):
+    _inherit = "stock.picking.in"
+    _columns = {
+        'fiscal_position': fields.many2one(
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think it's better not to store the fiscal position in stock.picking, and calculate it on the fly when making the invoice. You save all this code and a bit more.

rvalyi:
Need information

Hello @pedrobaeza @rdeheele for @renatonlima and me think it makes sense to store the fiscal_position in the picking. Indeed, you can very well create an invoice directly from a stock picking without any sale or purchase order, right? Then that picking should carry the proper operation to be able to copy it on the invoice just as @rdeheele did in the _prepare methods. In v8 a route would typically pre-fill the fiscal operation.

Now, conceptually, in Brazil, we do that in the localization made by Akretion, but we don't store the fiscal_position in the picking anymore. We store it instead at the stock move level because it can be different move by move in our country.

Would that be a valid implementation for you too @rdeheele ?
I'm concerned about accepting the field on the picking in that module as it could conflict with our usage. Eventually that could be an optional extra module. Please tell me first if having it at the stock move level (say for v8 we can contribute it) would be acceptable.

rvalyi:
Hum, correction, so in our localization, the fiscal position exists both at the picking and at the stock move level. By default the fiscal position of the move is copied from the picking unless it's manually set to something different.

@renatonlima could you confirm that code would be compatible for us too? (in v8 that would be the same unique picking object anymore).

account_fiscal_position_rule_stock/stock.py
```diff
@@ -75,3 +75,109 @@ def _prepare_invoice(self, cr, uid, picking, partner, inv_type,
...
             picking.fiscal_position.id
         return result
+
+
+class stock_picking_out(osv.Model):
+    _inherit = "stock.picking.out"
+    _columns = {
+        'fiscal_position': fields.many2one(
+            'account.fiscal.position', 'Fiscal Position',
+            domain="[('fiscal_operation_id','=',fiscal_operation_id)]"),
```
renatonlima: Where from this field fiscal_operation_id?

rvalyi: Indeed, @rdeheele I couldn't find that field either, could you tell us where does it come from? We kind of have that kind of "operation" thing here in Brazil, so eventually we can build synergies but first we need to understand where it comes from here. Isn't that from some of your custom modules?

rdeheele: @rvalyi @renatonlima it seems that this domain comes from a portugal/brazil localization, I find it here: https://github.com/proge/openerp-pt_br/blob/master/l10n_br_stock/stock.py#L12
I simply duplicate the existing code for in & out pickings, but it seems that this domain could be removed.

rvalyi: hehe this explains why! Ok fine then. For a minute we got some excitement to believe some core or OCA object would bring our localization closer. But OK, anyway we that kind of code coming into fiscal rules we are converging so please clean up and we evaluate again. We are pushing great stuff for v8, sorry for the delay but it's coming.

rdeheele:
At first sight, I just propose these changes to avoid the traceback https://github.com/OCA/account-fiscal-rule/issues/19
@pedrobaeza @rvalyi Generate fiscal position on the fly when making the invoice is smarter in terms of code, but see explicitly the fiscal position on picking may be clearer for the users.

pedrobaeza:
OK, no problem with me. If @rvalyi doesn't have any trouble, :+1:

lasley:
This one is pretty old with no activity and targeting v7, so I'm going to close. Please feel free to make suggested edits and reopen

PR #22, opened 2015-06-01
Feature/account fiscal position rule sale migration 8.0

renatonlima:

PR #25, opened 2015-07-28
Port account_fiscal_position_rule_sale to 8.0

bwrsandman:


account_fiscal_position_rule_sale/models/sale_order.py
```diff
@@ -23,82 +23,77 @@
...
-        if not kwargs.get('context', False):
-            kwargs['context'] = {}
-
-        kwargs['context'].update({'use_domain': ('use_sale', '=', True)})
-        obj_shop = self.pool.get('sale.shop').browse(
-            cr, uid, kwargs.get('shop_id'))
+    def _fiscal_position_map(self, cr, uid, result, shop_id=None,
+                             partner_id=None, partner_invoice_id=None,
+                             partner_shipping_id=None,
+                             company_id=None, context=None):
```
ghost: @bwrsandman you could use the api v8 here.

bwrsandman:
```
Element '<field name="shop_id">' cannot be located in parent view
```

Field doesn't exist in v8, see https://github.com/OCA/OpenUpgrade/blob/8.0/addons/sale/migrations/8.0.1.0/openupgrade_analysis_work.txt#L25

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Module name
+===========
+
+This module was written to allow you to include a rule to decide the correct
+fiscal position for Sale"
```
bwrsandman: quotes too many

account_fiscal_position_rule_sale/models/sale_order.py
bwrsandman: done

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Module name
```
pedrobaeza: Put the module name

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
...
+===========
+
+This module was written to allow you to include a rule to decide the correct
+fiscal position for Sale
```
pedrobaeza: s/Sale/sales orders

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
...
+
+Installation
```
pedrobaeza: Remove this section if empty

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
...
+============
+
+To install this module, you need to:
+
+* do this ...
+
+Configuration
```
pedrobaeza: Remove this section if empty

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
...
+=============
+
+To configure this module, you need to:
+
+* go to ...
+
+Usage
```
pedrobaeza: Remove this section if empty, although an explanation here is welcome

account_fiscal_position_rule_sale/README.rst
```diff
@@ -0,0 +1,77 @@
...
+
+To use this module, you need to:
+
+* go to ...
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
```
pedrobaeza: Remove this section if empty

account_fiscal_position_rule_sale/__openerp__.py
```diff
@@ -23,10 +23,8 @@
 
 {
     'name': 'Account Fiscal Position Rule Sale',
-    'version': '1.1.1',
+    'version': '1.1',
```
pedrobaeza: You can bump it to 1.0, as this is the first version for 8.0. According last discussions, this can be 8.0.1.0

account_fiscal_position_rule_sale/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
+# -*- encoding: utf-8 -*-
+###############################################################################
+#
+#   account_fiscal_position_rule_sale for OpenERP
+#   Copyright (C) 2009-TODAY Akretion <http://www.akretion.com>
```
pedrobaeza: `TODAY` is an abstract date that doesn't fit correctly I think. Better to put a fixed date.

account_fiscal_position_rule_sale/models/sale_order.py
```diff
@@ -0,0 +1,71 @@
...
+        return fp_rule.apply_fiscal_mapping(
+            result, company_id=company_id,
+        )
+
+    def onchange_partner_id(self, cr, uid, ids, partner_id, context=None):
+        result = super(SaleOrder, self).onchange_partner_id(
+            cr, uid, ids, partner_id, context=context)
+
+        values = result['value']
+        return self._fiscal_position_map(
```
pedrobaeza, line 49: I would prefer that this method only returns the corresponding fiscal position. Each method (onchange, write, create...) should adapt this result then to the corresponding needed data structure. This requires to change the `account_fiscal_position_rule`, but I think that the module will be much cleaner with this change.

pedrobaeza:
Tests for this feature are welcome

pedrobaeza:
Don't you think a better name for this should be `sale_fiscal_position_rule`?

account_fiscal_position_rule_sale/models/sale_order.py
bwrsandman, line 49: Could you do a pull request for that or at least write up an issue with what we can do for this?

How about making this the 8.0.2.0 version?

bwrsandman:
The name change isn't worth the hassle of renaming all the depend modules and writing migration scripts.

But the name is good the way it is:
The main module `fiscal_position_rule` -> first
The secondary module is `sale` -> second

The name says: this is the `fiscal_position_rule` module for dealing with `sale` module

pedrobaeza:
The problem I see is the word account_ in first place, that breaks current logic (if any) of naming conventions. As this one is not a dependency for any other module, I see no special problem in the renaming.

bwrsandman:
@pedrobaeza it is a dependency for the quebec and canadian pay modules

pedrobaeza:
But both are going to be ported also, so you can change the dependency name in the porting process.

bwrsandman:
It makes OpenUpgrade scripts more complicated when naming (to something I don't think is very necessary) is changed.

pedrobaeza:
Well, it's only 1 line script change (in the file apriori.py), but I don't want you to go discussing this, so I leave my pretension.

danimaribeiro:
I'm starting to test this, I've created this general rule to get all the cases.
general rule

But this rule is not working as expected, when I change the partner it should return the fiscal position set in the rule, but nothing happens.

I think there are some problems when passing partner_id and company_id to fiscal_position_map:
Both parameters are None here.
https://github.com/OCA/account-fiscal-rule/blob/8.0/account_fiscal_position_rule/account_fiscal_position_rule.py#L121

account_fiscal_position_rule_sale/models/sale_order.py
danimaribeiro, line 49: Another good thing to do, is return the rule ID together with the fiscal position.
When there are a lot of rules, is good to know the rule was applied.

rvalyi:
Hello, I would say superseded by https://github.com/OCA/account-fiscal-rule/pull/22 now merged. Also the PR #22 was from 2014. I'm kind of surprised somebody worked again on that instead of fixing the previous PR... So  propose to close that one, but if we missed something please tell us.

PR #27, opened 2015-07-28
Port account_fiscal_position_rule_stock to 8.0

bwrsandman:


account_fiscal_position_rule_stock/README.rst
```diff
@@ -0,0 +1,74 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Module name
```
pedrobaeza: Same comments as for `account_fiscal_position_rule_sale` for the manifest

account_fiscal_position_rule_stock/models/stock_picking.py
```diff
@@ -0,0 +1,83 @@
+# -*- encoding: utf-8 -*-
+###############################################################################
+#
+#   account_fiscal_position_rule_stock for OpenERP
+#   Copyright (C) 2009-TODAY Akretion <http://www.akretion.com>
```
pedrobaeza: Change TODAY reference

account_fiscal_position_rule_stock/models/stock_picking.py
```diff
@@ -0,0 +1,83 @@
...
+    _inherit = "stock.picking"
+
+    fiscal_position = fields.Many2one(
+        'account.fiscal.position',
+        string='Fiscal Position',
+        domain="[('fiscal_operation_id','=',fiscal_operation_id)]",
+    )
+
+    @api.model
+    def _fiscal_position_map(
```
pedrobaeza, line 35: Same argument for this method that in sale

account_fiscal_position_rule_stock/models/stock_picking.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+        return fp_rule_obj.apply_fiscal_mapping(
+            result,
+            partner_id=partner_id,
+            partner_invoice_id=partner_invoice_id,
+            partner_shipping_id=partner_shipping_id,
+            company_id=company_id,
+        )
+
+    def onchange_partner_in(self, cr, uid, ids, partner_id=None,
```
pedrobaeza, line 50: onchange_partner_in? Shouldn't it be `onchange_partner_id`?

account_fiscal_position_rule_stock/models/stock_picking.py
```diff
@@ -0,0 +1,83 @@
...
+            company_id=company_id,
+            context=context,
+        )
+
+    @api.model
+    def _prepare_invoice(self, picking, partner, inv_type, journal_id):
+        result = super(StockPicking, self)._prepare_invoice(
+            picking, partner, inv_type, journal_id
+        )
+        result['fiscal_position'] = (
```
pedrobaeza: This can be with new API: `result['fiscal_position'] = picking.fiscal_position.id`, but this shouldn't overwrite the current value if there's no fiscal position, so it should be:

```
if picking.fiscal_position:
    result['fiscal_position'] = picking.fiscal_position.id
```

pedrobaeza:
I see that this module lacks the propagation of the fiscal position from the sale order to the picking, and tests that exercises these changes.

account_fiscal_position_rule_stock/models/stock_picking.py
bwrsandman, line 50: You'd think that, but no

https://github.com/odoo/odoo/blob/a9e3d74713b6882c64617345696ad326d75a4745/addons/warning/warning.py#L158

pedrobaeza, line 50: Curious...

danimaribeiro, line 50: "onchange_partner_in" is defined only in the module warning, if I don't have installed, it gives me an error: "AttributeError: 'super' object has no attribute 'onchange_partner_in'"

account_fiscal_position_rule_stock/models/stock_picking.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class StockPicking(models.Model):
+    _inherit = "stock.picking"
+
+    fiscal_position = fields.Many2one(
+        'account.fiscal.position',
+        string='Fiscal Position',
+        domain="[('fiscal_operation_id','=',fiscal_operation_id)]",
+    )
```
danimaribeiro, line 32: 'fiscal_operation_id' does not exists nowhere.

danimaribeiro:
@bwrsandman Are you going to work on these changes? if not I can fix these problems.

@pedrobaeza  About the propagation of fiscal position, do you think this module is the best place to do it? 
I have some tests in the brazilian localization here that test this propagation:
https://github.com/odoo-brazil/l10n-brazil/blob/8.0/l10n_br_sale_stock/test/sale_order_demo.yml

Also I can work on some tests for the whole module.
I just don't know the best way to handle the onchange issue. Put the warning module as dependency?

bwrsandman:
@danimaribeiro Please do. I am no longer doing odoo and I don't have the free time to work on these.

PR #29, opened 2015-09-15
fixed tests: relying on ir_model_data record update was insufficent to…

rvalyi:
… set test data consistently after module install

### DIFF
--- a/account_fiscal_position_rule/test/test_rules.yml
+++ b/account_fiscal_position_rule/test/test_rules.yml
@@ -55,7 +55,7 @@
       'partner_id': ref('base.res_partner_11'),
       'partner_invoice_id': ref('base.res_partner_11')
     }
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

account_fiscal_position_rule/test/test_rules.yml
```diff
@@ -65,6 +55,7 @@
       'partner_id': ref('base.res_partner_11'),
       'partner_invoice_id': ref('base.res_partner_11')
     }
+    self.env['res.partner'].browse(ref('base.main_company')).write({'country_id': ref('base.us')})
```
renatonlima: Do browse res.partner with ref('base.main_company')? the correct reference should be ref('base.main_partner')?

rvalyi:
@renatonlima good catch, you are right! It worked by luck because both ids are 1 by default. I'm fixing this.

renatonlima:
:+1:  LGTM

PR #21, opened 2015-06-01
Feature/account fiscal position rule stock migration 8.0

renatonlima:

PR #30, opened 2015-09-26
Port account_fiscal_position_rule_stock to 8.0 continuation to #27

danimaribeiro:
Continuation of this PR https://github.com/OCA/account-fiscal-rule/pull/27
Added some unit tests, added propagation of the fiscal position from the sale to picking.
@bwrsandman 

@renatonlima and @mileo


mileo:
Testing in http://3115974-30-5b3115.runbot1.odoo-community.org/web#id=24&view_type=form&model=stock.picking&action=316&active_id=2

OUT/00007 map stock fiscal position, but the fiscal postion isn't copy to the invoice.

https://github.com/OCA/account-fiscal-rule/pull/30/files#diff-473eef115068a06e3800748db81bb2beR73 

please use get_invoice_vals https://github.com/odoo-brazil/account-fiscal-rule/blob/8.0/account_fiscal_position_rule_stock/stock.py#L57 instead of prepare_invoice_vals 

https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=_get_invoice_vals

mileo:
Just chery-pick commit https://github.com/odoo-brazil/account-fiscal-rule/commit/3eaadf53fd029420f1ed17490b1aa91507532443

rvalyi:
also there has been some kind of "double KO" on that PR again, we should ensure we cherry-pick any improvement from this branch https://github.com/OCA/account-fiscal-rule/pull/21 from @renatonlima if we merge this one. I didn't check yet sorry.

rvalyi:
Also in fact if PR #21 does the job, it has the anteriority and I would find it not normal we merge that later one, that's the problem of the author if the missed the previous PR I would say. So :-1: for me at this stage but I'll analyse things further.

mileo:
@rvalyi #21 is missing 3eaadf53fd029420f1ed17490b1aa91507532443 too

danimaribeiro:
#17  Has even higher priority. What we do now? Since march waiting.
You can chose #21 to keep your commits, but I won't help in this repo anymore.

We wait too long for the others PR to be accepted, I chose to go with the other code because they also migrate the other modules and they were active helping and reviewing the code. I think the guys from Canada were also fed up waiting to some comments in the migration and decide to continue on their own.

danimaribeiro:
@mileo I'll get those changes, and add some tests to it.

danimaribeiro:
@mileo Problems fixed, added unit tests to that method.

max3903:
@danimaribeiro Can you rebase and fix the conflict please?

danimaribeiro:
Done

rvalyi:
@danimaribeiro may be we can use this branch, I'm not sure. But you said "I think the guys from Canada were also fed up waiting to some comments". Which comments please? https://github.com/OCA/account-fiscal-rule/pull/17 and https://github.com/OCA/account-fiscal-rule/pull/21 had simply no review at all as far as I can see.

As for the errors we had in the Travis build in this repo, there appeared after a change in the OCA Travis management, they were not due to these branches (and BTW I fixed them because nobody fixed them during this period).

So for me Sandy should just have look in the existing PR's. They might not have been perfect at that time, but they were at least close. Anyway, I'll try to look at these two branches tomorrow.

max3903:
:+1:

rvalyi:
:-1: 
This isn't too bad but there is something we really don't like: the dependency to the warning module: it's a terribly bad module we never use and that is totally non-modular (IMHO a module Odoo SA should just trash). It would also force us to install the sale and purchase modules while possibly you only want account and stock. That kills modular implementations by steps and makes migrations harder...

So with @renatonlima we just reviewed that and our conclusion is:
- please remove the warning module dependency
- then don't call super in the on_change_partner_in or at least test if it is defined before calling it.
- we would prefer to name the on_change method name to on_change_partner_id (feels more natural then)
- please add the stock_account module as a dependency. @renatonlima suspects it would get picked by the auto_install flag but installed only at the end. But here you need it in the view for the invoice_type picking field for instance. It may break if you install the account_fiscal_position_rule_stock alone (while it may pass on Travis because Travis install all the branch modules together). We aren't sure about this last point, but that costs nothing or at least should be tested.

if one really want to install account_fiscal_position_rule_stock along with the warning module then he should make a new extra "glue" module to support both, but in any case we shouldn't be forced to have this warning module by default.

Otherwise looks good, specially the test part.

rvalyi:
cc @danimaribeiro @mileo @max3903

mileo:
@oca Can some one fix runbot dependencies?

rvalyi:
@mileo @oca seems it's because the runbot is not using https://github.com/Trust-Code/account-fiscal-rule/blob/8.0_account_fiscal_position_rule_stock/oca_dependencies.txt
I have no idea why. Fixing it would be cool but I don't think that's a blocker here, we often merge stuff without having Runbot green...

mileo:
I would like to test on it, but do some local testing in a few minutes

danimaribeiro:
Any other idea to solve the problem with the warning, that can make then work together without an extra module?

If not, we should at least create this module here to fix that.

rvalyi:
@danimaribeiro I see two possibilities:
1. either test if method is defined before calling super in case it's warning is installed too. If you can do that, it's the simplest ( I think you can go with https://docs.python.org/2/library/functions.html#hasattr )
2. either add an extra module

In any case we prefer not installing this module but then that's may be a matter of taste.

rvalyi:
I think @danimaribeiro you'll also come to the conclusion that some official Odoo SA modules are simply better not installed. Typically such a module is so crappy that in v10 it may be removed and then you are screwed anyway. Doing that helped us to stay away from the wiki, the ftp, the old 6.\* e-commerce modules or I would even say now from the website modules... At least in the past we got it right...

rvalyi:
@danimaribeiro Well in fact I'm not sure if the hasattr method would work. because if warning is installed last it may destroy our on_change method in the view definition no matter what. So I would say an extra-module then. If you are courageous why not. On our side as I said we really don't care about the warning module and the priority is really to get that account_fiscal_position_rule_stock module in a first step so it makes it easier to finalize the OCA merges in l10n-brazil. After that if you want to propose some module to support the warning module that's fine.

danimaribeiro:
Ok, another module is the way to go.

rvalyi:
@danimaribeiro but if you could just remove the warning dependency here 1st that would be cool. We already got 1 approval, if @mileo and @renatonlima approve we get the merge in minutes. And as @renatonlima is working right now to get the l10n_br_stock_account green in l10n-brazil, that would help...

mileo:
:+1:  LGTM

renatonlima:
:+1:

rvalyi:
thanks!

rvalyi:
@danimaribeiro gah, we forgot to update the list of migrated modules in the README.rst
If you could do that in another PR...

mileo:
@danimaribeiro Can you make a PR to 9.0 too?

PR #31, opened 2015-10-01
fixed __init__.py import

rvalyi:
oups just screwed the files layout re-organization missing an import, this was meant to be part of the previous commit.


rvalyi:
I mean part of this one: https://github.com/OCA/account-fiscal-rule/pull/22

PR #33, opened 2015-11-03
[FIX] invalid name fields taxes

renatonlima:


rvalyi:
:+1: 

@legalsylvain  @renatonlima is right: 
indeed in the product.template, the fields are called 'supplier_tax_ids' and 'customer_tax_ids'. But here this is the case somebody writes into the the fiscal classification and in this objects the fields are called:
'purchase_tax_ids' and 'sale_tax_ids'. So the detection should use these fields. Then the write method inside the product.template will take care of propagating the right taxes in the templates using that classification.

ping @mileo @danimar @danimaribeiro please approve if you agree as we need that one to have the tests pass for https://github.com/OCA/l10n-brazil/pull/134 and we are nearly good otherwise. So it can take less than 5 days if we have 3 approvals. Let's have that to get enough merges for the Friday meeting...

legalsylvain:
:+1: Thanks !

@renatonlima : I can not figure out why runbot fails on dependencies. (server-tools / web_context_tunnel). (Configuration seems to be good [here](https://github.com/akretion/account-fiscal-rule/blob/akretion-team/fix-fiscal-classification-taxes/oca_dependencies.txt) ...)

danimaribeiro:
:+1:

mileo:
:+1:

PR #34, opened 2015-11-03
[FIX] invalid name fields taxes

renatonlima:


rvalyi:
@renatonlima I think v9 will have automatic and systematic forward porting for now. At least I have seen no OCA project manually porting v8 fixes to unmigrated v9 modules so far, so I propose to close that one for now.

PR #35, opened 2015-11-04
Akretion team/fiscal classification template

renatonlima:
back fiscal classification template model


account_product_fiscal_classification/model/account_product_fiscal_classification.py
```diff
@@ -93,7 +126,7 @@ def _compute_product_tmpl_info(self):
     def write(self, vals):
         res = super(AccountProductFiscalClassification, self).write(vals)
         pt_obj = self.env['product.template']
-        if 'supplier_tax_ids' in vals or 'customer_tax_ids' in vals:
+        if 'purchase_tax_ids' in vals or 'sale_tax_ids' in vals:
```
legalsylvain, line 84: It has been fixed in an other PR #33 . I guess you should undo this changes and rebase your PR for more visibility.

account_product_fiscal_classification/model/account_product_fiscal_classification.py
```diff
@@ -44,6 +41,47 @@ def _default_company_id(self):
 
     description = fields.Text()
 
+    active = fields.Boolean(
+        default=True,
+        help="If unchecked, it will allow you to hide the Fiscal"
+        " Classification without removing it.")
+
+
+class AccountProductFiscalClassificationTemplate(models.Model):
```
legalsylvain: Could you set a file per model ? 
- OCA Convention 
- Code review will be more easy to do : For the time being github diff display that AccountProductFiscalClassification disappeared, replaced by AccountProductFiscalClassificationModel.

rvalyi:
:+1:  Okay for me except please increment the version from 8.0.2.0.0 to 8.0.2.1.0 as significant feature is added and showing up only if you restart with upgrade.
see guideline: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

mileo:
:+1:

rvalyi:
cc @legalsylvain

rvalyi:
we've got 2 approvals and the 5 calendar days. @renatonlima is working on some more l10n-brazil testing. If everything goes green (as it seems it will), I'll merge that in a few hours.

account_product_fiscal_classification/tests/tests.py
```diff
@@ -138,3 +144,12 @@ def test_06_unlink_fiscal_classification(self):
             [self.classification_1_id])
         with self.assertRaises(ValidationError):
             classification.unlink()
+
+    def test_07_classification_generate_from_template(self):
```
danimaribeiro, line 148: This test seems to not be working:
https://coveralls.io/builds/4063409/source?filename=account_product_fiscal_classification%2Fmodels%2Fwizard_account_product_fiscal_classification.py

rvalyi, line 148: @danimaribeiro good catch.
Now does anyone understands why this test doesn't run? I'm not getting it...

rvalyi, line 148: @danimaribeiro @legalsylvain 
In fact it seems we have some bad issue here: it seems that no test is being run at all in this module.
So look at test_01 for instance. It's supposed to call wizard method button_change_fiscal_classification:
https://github.com/akretion/account-fiscal-rule/blob/akretion-team/fiscal-classification-template/account_product_fiscal_classification/tests/tests.py#L63
but it doesn't happen!
https://coveralls.io/builds/4063409/source?filename=account_product_fiscal_classification%2Fmodels%2Fwizard_change_fiscal_classification.py#L48

Worse: the test didn't run either on last patch by @renatonlima on Nombember 3rd:
https://coveralls.io/builds/4037079/source?filename=account_product_fiscal_classification%2Fmodel%2Fwizard_change_fiscal_classification.py#L48

I'm now wondering if we merged that module with tests simply disabled....

cc @mileo

rvalyi, line 148: so in case somebody is looking at that, I wondered if eventually the Odoo TransactionCase test would run at install and not later at update, so I tried this change:
https://github.com/OCA/account-fiscal-rule/commit/1cbc107b75f6f79cb2e72e28928e11cfd2a986e4
Sadly that didn't change anything according to the coveralls results

pedrobaeza, line 148: One silly thing: try to rename tests.py to something else: test_account_product_fiscal_position.py for example. Are these tests run locally?

danimaribeiro, line 148: Or probably something related to the class name:
https://github.com/akretion/account-fiscal-rule/blob/akretion-team/fiscal-classification-template/account_product_fiscal_classification/tests/tests.py#L27

"TestSomething" instead of "Tests"

rvalyi, line 148: @pedrobaeza it happens that I'm currently working on a very poor hardware these days and looking at Travis directly is easier for me, well at least was until I understood it wasn't trivial. Trying a rename now..

rvalyi, line 148: @pedrobaeza @danimaribeiro this one is driving me crazy. I tried both suggestions and tests keep not running:
https://github.com/akretion/account-fiscal-rule/commit/f61bd3ae037ba1e3b1daca2e2974c85b278889d5
https://github.com/akretion/account-fiscal-rule/commit/62cbb6cc04e11b8c64d8614b19d0ee11e8a67412

I'm testing this on my 2Gb i3 slow disk super laptop then now... (got a laptop crash)

danimaribeiro:
Also, I dont see import tests in init.py

rvalyi:
so...

my previous attempt failed because I screwed the tests/**init**.py and I had a false Travis positive while import failed.

Anyway, it turns out that simply renaming tests/tests.py to tests/test_classification.py for instance magically turns the tests on...

Now the bad news is that the tests are badly broken... in fact I tried that on @legalsylvain original module before @renatonlima template refactor and the tests are badly broken already.

It shouldn't be that bad as our local tests with the l10n-brazil localization seem to all pass smoothly. So probably it's just the tests of this module that weren't properly adapted during @legalsylvain refactor to try to match the older Akretion module and wrongly assume to pass when they failed silently; only guessing here.

I'll push some branch making the problem explicit to all and try to work out a solution. Thank you @pedrobaeza  for suggesting that name change, it would have take me an absurd amount of time before I would have tried that renaming by luck.

conclusion about Python mantra: #explicitisbettertthanimplicitmyass

rvalyi:
proof here: https://github.com/OCA/account-fiscal-rule/pull/37

rvalyi:
So I propose to fix upstream tests in https://github.com/OCA/account-fiscal-rule/pull/37 and then rebase this PR once tests are fixed so we deal with one problems one by one.

rvalyi:
Needless to say, contribs welcome to make the tests pass.

rvalyi:
Arghhhh!!
so I just tried to squash 6 commits from @renatonlima as suggested by Pedro in https://github.com/OCA/account-fiscal-rule/pull/38 but I screwed the rebase making this impossible to merge. My plan was fixing the tests first and then rebase that branch upon the fix, but that just failed.

So let's say this is now superseeded by https://github.com/OCA/account-fiscal-rule/pull/38

PR #37, opened 2015-11-11
magically turn testing on by renaming tests/tests.py to tests/test_cl…

rvalyi:
…assification.py; tests are broken


rvalyi:
@legalsylvain @renatonlima @mileo @danimaribeiro @pedrobaeza 

So here is at least a branch starting right on @legalsylvain refactor/port and showing that tests were simply disabled because of of their name "tests.py" and that once enabled they were failing badly:
https://travis-ci.org/OCA/account-fiscal-rule/jobs/90452122

rvalyi:
I could get most of the tests working again by fixing a trivial xml_id ref error in https://github.com/OCA/account-fiscal-rule/commit/91451ec935b652e6f5a16ec75e34faa361d3f54d

Now it seems we have more serious errors to fix and as it's 1h44 AM, you'll excuse me I'll leave that battle for tomorrow.

rvalyi:
contribs welcome to make the tests pass

rvalyi:
superseeded and merged

PR #38, opened 2015-11-11
Test pass

mileo:
FIX #37 #35 

Good night!


mileo:
Any one knows why we have this message, the oca_dependencies.txt seems to be ok

File "/srv/openerp/instances/openerp-oca-runbot/parts/odoo-extra/runbot/static/build/3119407-38-da5905/openerp/addons/base/module/module.py", line 393, in state_update
    raise orm.except_orm(_('Error'), _("You try to install module '%s' that depends on module '%s'.\nBut the latter module is not available in your system.") % (module.name, dep.name,))
except_orm: ('Error', u"You try to install module 'account_fiscal_position_rule_purchase' that depends on module 'web_context_tunnel'.\nBut the latter module is not available in your system.")

pedrobaeza:
Can you squash the commit history a bit?

I have added the dependency to runbot and restart the build.

mileo:
@pedrobaeza Done

pedrobaeza:
I still see 17 commits with very weak modifications like pep8 and so on. Please join also that.

mileo:
@pedrobaeza You are talking about the Akretion's commits? cc @renatonlima @rvalyi

rvalyi:
@pedrobaeza @mileo I'll squash the commits.
One reason why we broke down the commits so much is because many people are working on this in parallel or depending on this for further developments so we shared anything we could as quickly as possible. But no problem we can now squash some commits.

rvalyi:
superseeded by https://github.com/OCA/account-fiscal-rule/pull/39 where I squashed 5 commits in @renatonlima work.

PR #40, opened 2015-11-18
fix dependency in module account_fiscal_position_rule_stock change st…

renatonlima:
After version 8.0 inventory module was divided into two stock and stock_account , the stock_account module is installed automatically because it has the autoinstall attribute, but I think important to make the correct dependence explicit , you can prevent future problems and leave the more readable code.


max3903:
:+1:

rvalyi:
:+1: LGTM

renatonlima:
cc @mileo @danimaribeiro

danimaribeiro:
:+1:

rvalyi:
3 votes, merging...

PR #45, opened 2015-12-04
Akretion/9.0 migration account fiscal rule

mbcosta:


account_fiscal_position_rule/views/account_fiscal_position_rule_view.xml
```diff
@@ -1,220 +1,231 @@
...
+				  parent="account.menu_finance_configuration" name="Fiscal Position Rule" sequence="21"/>
+
+		<!-- Fiscal Position Rule Template -->
+		<record id="view_account_fiscal_position_rule_template_form" model="ir.ui.view">
+			<field name="name">account.fiscal.position.rule.template.form</field>
+			<field name="model">account.fiscal.position.rule.template</field>
+			<field name="arch" type="xml">
+				<form string="Fiscal Position Rule Template">
+					<group>
+						<group string="General" colspan="4">
```
danimaribeiro: I suggest put the name attribute in the groups, in version 9 we can't use the string attribute to inherit views anymore.

mbcosta: Hi danimaribeiro, thanks for the feedbak, I add the attribute name

rvalyi:
@mbcosta: just a hint in case you don't know it: when you mention somebody in a comment, if you use @ like @danimaribeiro then you are sure that person will be notified. Else it depends a bit if the person is following the thread or not. The fix makes sense BTW.

mbcosta:
Thanks for the hint @rvalyi

account_fiscal_position_rule/models/account_fiscal_position_rule.py
```diff
@@ -116,52 +115,47 @@ def _map_domain(self, partner, addrs, company, **kwargs):
...
+            fsc_pos = self.search(domain, limit=1)
+            if fsc_pos:
+                result = fsc_pos[0].fiscal_position_id
 
         return result
 
-    def apply_fiscal_mapping(self, result, **kwargs):
-        result['value'].update(self.fiscal_position_map(**kwargs))
+    def apply_fiscal_mapping(self, **kwargs):
+        result = self.fiscal_position_map(**kwargs)
```
rvalyi: better to write simply return self.fiscal_position_map(**kwargs)

rvalyi: please use simply return self.fiscal_position_map(**kwargs)

rvalyi (changes requested):

rvalyi:
:+1:

renatonlima:
:+1:

PR #48, opened 2015-12-08
Akretion/9.0 migration account prod fiscal posit stock

mbcosta:


lasley:
Thanks for the submission @mbcosta. I was about to review this one and noticed the age + that tests have been failing for some time. Did you plan on continuing this?

mbcosta:
Hi @lasley , thanks for review
   Well it's long time ago, in this PR I just did "superficial" migration, the error is refer to missed @api.model where the original code still use the old api. But the real problem here is that the function to create invoices from pickings was removed since 9.0 from core, what made this module useless, in version 10.0 @rvalyi make a PR to reimplent it https://github.com/OCA/account-invoicing/pull/320 ( this function is very important in brazilian localization ) now we need to check if is worth to keep this PR open and make an backport of the module or just close and work only in version 10.0, by the time this PR are open it's seem better go to 10.0, what do you think ?  

cc @renatonlima @rvalyi

lasley:
@mbcosta IMO just hopping to v10 would be best. v8 is EoL, so adding more development effort into it is futile unless you have a specific reason. 

I'll close this then. Feel free to reopen if you decide it's best to go this route.

PR #50, opened 2016-01-12
[FIX] Update product_template taxes

mileo:
Select a fiscal classification linked to multiple products templates, change taxes and try to save it. 

You get this error: 

ValueError

Expected singleton: product.template(9, 11, 37, 38, 6, 18, 48, 30, 26, 46, 21, 19, 20, 34, 35, 10, 40, 8, 5, 13, 29, 27, 28, 23, 22, 12, 14, 33, 44, 3, 2, 7, 4, 31, 32, 39, 25, 24, 16, 17, 15, 47, 1, 49, 41, 50, 36, 42, 43, 45)

### DIFF
--- a/account_product_fiscal_classification/models/account_product_fiscal_classification.py
+++ b/account_product_fiscal_classification/models/account_product_fiscal_classification.py
@@ -84,7 +84,8 @@ class AccountProductFiscalClassification(models.Model):
         if 'purchase_tax_ids' in vals or 'sale_tax_ids' in vals:
             for fc in self:
                 pt_lst = pt_obj.browse([x.id for x in fc.product_tmpl_ids])
-                pt_lst.write({'fiscal_classification_id': fc.id})
+                if pt_lst:
+                    pt_lst.write({'fiscal_classification_id': fc.id})
         return res
 
     @api.multi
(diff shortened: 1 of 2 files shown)


account_product_fiscal_classification/models/account_product_fiscal_classification.py
```diff
@@ -84,7 +84,8 @@ def write(self, vals):
         if 'purchase_tax_ids' in vals or 'sale_tax_ids' in vals:
             for fc in self:
                 pt_lst = pt_obj.browse([x.id for x in fc.product_tmpl_ids])
-                pt_lst.write({'fiscal_classification_id': fc.id})
+                if pt_lst:
```
pedrobaeza, line 87: This is not needed, unless you have put a `self.ensure_one()` in the write method, so you should remove that part instead of patching this call (and every possible call). `write` method (and overrides) should accept 0 or N records.

rvalyi:
@mileo I agree with your fix, but could you just take @pedrobaeza remark into account so we can approve it?

mileo:
Yes, my fault. I forgot to correct this problem.

rvalyi:
up @mileo any time to fix one this so we can merge and close it?

rvalyi:
@mileo in case you have any time for that one during the sprints...

account_product_fiscal_classification/models/account_product_fiscal_classification.py
leorochael: In other words, read the traceback of the "Expected Singleton" error to find out who overwrote the `write()` method of `product.template` and added a `self.ensure_one()`, and find a way to remove that `self.ensure_one()` call instead of adding an `if` here.

mileo:
Hello I can not repeat the problem. So I'm closing the PR.

rvalyi:
ok thanks for having checked that.

PR #54, opened 2016-09-04
9.0 classification per category

legalsylvain:
Add possibility to restrict fiscal settings on product, depending of its categories.

This is usefull to reduce errors, and make more easy fiscal settings on products. French example : in a food shop. you can have  5.5% VAT or 20% VAT.
But if the product is in the "alcool / beer" category so only 20% VAT is a correct setting. 

category_with_fiscal_restriction


legalsylvain:
CC : @sebastienbeau, @rvalyi, @danimaribeiro.

rvalyi:
@legalsylvain thanks for this long wanted feature!  i like it but please you should make the tests pass https://travis-ci.org/OCA/account-fiscal-rule/jobs/160804830#L708
cc @renatonlima

legalsylvain:
@rvalyi : Thanks for your review. I fixed the problem. (sorry for the late ! I totally forgot this PR)

@sebastienbeau, @bealdav, @alexis-via  : Any PoV ? I really see this feature as a good nice to have in France.

Thanks for your review.

rvalyi:
:+1: @legalsylvain looks good. One question/suggestion: suppose I set only one fiscal category allowed on some category. Then may be when creating the product it should come with that fiscal category by default, no? In any case it would just save a click and not make a huge usability difference...

legalsylvain:
@rvalyi. The feature you describe is implemented. Here : 
https://github.com/OCA/account-fiscal-rule/pull/54/files#diff-427f6fda9e41d700a0d351c6fbf9f432R61

@all : Extra fix in the last commit. (parent_id has been removed in V9.0 on account.tax and account.tax.template.)

Thanks for your review.
CC : @mileo, @renatonlima

rvalyi:
@legalsylvain thanks for pointing the feature. So again :+1: for me

legalsylvain:
Thanks @rvalyi for your "again :+1:". Unfortunately, not sure it counts for 2 :+1: ;-) This PSC seems to be quiet empty, no answer except you ! Let wait...
Regards.

rvalyi:
@legalsylvain at least @renatonlima will review soon too and possibly @mbcosta. In any case we should fix the official PSC of this repo: Renato is the main committer/author but isn't a repo PSC, I'm the second committer and I have commit rights because I have global commit rights without being explicitly a repo PSC either. Eventually we will candidate officially to fix this, specially before v10 migration.

jweste:
+1 for me too, thanks!

renatonlima:
Hi guys!

:+1: This feature also can be nice have to for brazilian configuration

PR #49, opened 2015-12-08
Akretion/9.0 migration account prod fiscal posit purchase

mbcosta:

PR #47, opened 2015-12-08
Akretion/9.0 migration account prod fiscal posit sale

mbcosta:

PR #56, opened 2016-11-14
[9.0][FIX] Return object instead dict when Fiscal Position is False

mbcosta:
When Fiscal Position not found the method need to return a Object instead Dictionary. Besides the fix I removed parameters not used in new API ( cr, uid and context ) and replace Tabs to Space in Data File.

The Migrations PRs  https://github.com/OCA/account-fiscal-rule/pull/47 https://github.com/OCA/account-fiscal-rule/pull/48 https://github.com/OCA/account-fiscal-rule/pull/49 depends of this fix to pass the tests.

### DIFF
--- a/account_fiscal_position_rule/models/account_fiscal_position_rule.py
+++ b/account_fiscal_position_rule/models/account_fiscal_position_rule.py
@@ -57,6 +57,7 @@ class AccountFiscalPositionRule(models.Model):
         self.from_country = self.company_id.country_id
         self.from_state = self.company_id.state_id
 
+    @api.multi
     def _map_domain(self, partner, addrs, company, **kwargs):
         from_country = company.partner_id.country_id.id
         from_state = company.partner_id.state_id.id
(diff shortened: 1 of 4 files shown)


account_fiscal_position_rule/models/account_fiscal_position_rule.py
```diff
@@ -192,8 +192,7 @@ class WizardAccountFiscalPositionRule(models.TransientModel):
         default=lambda self: self.env['res.company']._company_default_get(
             'wizard.account.fiscal.position.rule'))
 
-    def _template_vals(self, cr, uid, template, company_id,
-                       fiscal_position_id, context=None):
+    def _template_vals(self, template, company_id, fiscal_position_id):
```
pedrobaeza, line 199: Put `@api.multi` below

rvalyi:
:+1:

mbcosta:
Hi @pedrobaeza thanks for review, I included decorator on methods.

pedrobaeza (approved):

rvalyi:
:+1:

renatonlima (approved):
:+1:

PR #62, opened 2016-12-12
Migrate account_product_fiscal_classification to V10

yk2kus:


yk2kus:
I need help with the syntax error
https://goo.gl/QZypwc

rvalyi:
@yk2kus I'm not very sure, but for the rst error, you can try to remove 2 spaces at lines 22 and 23 of the README.rst file. If it doesn't do the trick, you can also try to insert a blank line before line 22, but only if removing space alone doesn't fix the syntax problem.

renatonlima:
:+1:

account_product_fiscal_classification/models/wizard_change_fiscal_classification.py
```diff
@@ -27,11 +27,14 @@ def _default_old_fiscal_classification_id(self):
         required=True, domain="[('id', '!=', old_fiscal_classification_id)]")
 
     # View Section
-    @api.one
```
rvalyi, line 30: :-1: sorry if I'm wrong but shouldn't we have an @api.multi decorator here then?

yk2kus: @rvalyi not required but can be placed. It will not have any impact in the result because old api is not supported in V10, also will not raise any issue because we can loop through a single recordset as well. I would like to know if it will make an impact ?

rvalyi: In fact, I think api.one is not deprecated anymore, look at all the compute methods in account/models/account_invoice.py for instance, they all use the @api.one formalism. Can we not switch back to @api.one and the former impl assuming self is the record?
see for instance https://github.com/odoo/odoo/blob/10.0/addons/account/models/account_invoice.py#L88

yk2kus:
From what I can understand, it is deprecated but not discontinued. Source https://github.com/odoo/odoo/issues/12408#issuecomment-226421501 . My purpose of removal was lint check https://goo.gl/CVVnks

account_product_fiscal_classification/models/wizard_change_fiscal_classification.py
yk2kus: From what I can understand, it is deprecated but not discontinued. Source https://github.com/odoo/odoo/issues/12408#issuecomment-226421501 . My purpose of removal was lint check https://goo.gl/CVVnks , should we still switch back?

yk2kus: @rvalyi waiting for your response.

lasley: @yk2kus - @rvalyi is correct that `api.one` is deprecated and should not be used in new code.

You can either alter to `api.multi` or switch to [oca_decorators](https://github.com/OCA/oca-decorators), which has a [`foreach`](https://github.com/OCA/oca-decorators/blob/master/oca/decorators/foreach.py#L9) to replace the deprecated `.one`

account_product_fiscal_classification/demo/product_category.xml
```diff
@@ -5,12 +5,12 @@
     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 -->
 
-<openerp><data>
+<odoo><data>
```
lasley, line 8: Please remove the `data` tags - they are implied with the new `odoo` tag

account_product_fiscal_classification/models/account_product_fiscal_classification.py
```diff
@@ -47,13 +47,13 @@ def _default_company_id(self):
...
     # Compute Section
-    @api.one
     def _compute_product_tmpl_info(self):
-        res = self.env['product.template'].search([
-            ('fiscal_classification_id', '=', self.id), '|',
-            ('active', '=', False), ('active', '=', True)])
-        self.product_tmpl_ids = res
-        self.product_tmpl_qty = len(res)
+        for record in self:
+            res = self.env['product.template'].search([
```
lasley, line 52: Use `.with_context(active_test=False)` instead of the `|` on `active`

account_product_fiscal_classification/models/account_product_fiscal_classification_model.py
```diff
@@ -21,7 +21,7 @@
 ##############################################################################
```
lasley, line 21: Please switch long headers to [shortened header style](https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3)

account_product_fiscal_classification/models/product_template.py
```diff
@@ -55,35 +55,43 @@ def _check_classification_categ(self):
                     template.categ_id.complete_name,
                     ''.join(
```
lasley, line 56: `''.join(template.categ_id.fiscal_classification_ids.mapped('name'))`

lasley (changes requested):
Thanks @yk2kus - code, no test

rvalyi:
guys, I'm closing this PR as it has been superseded by https://github.com/OCA/account-fiscal-rule/pull/74 which has been merged.

PR #64, opened 2016-12-20
[9.0][FIX] account_fiscal_position_rule_sale: Fix correct fiscal position when apply a rule

sergio-teruel:
This PR assure the correct order when onchange method is called.
ping @rvalyi 
cc @tecnativa


pedrobaeza (approved):

yajo (approved):

Armitxes:
This PR might fix the issue for sale but not for purchase.
Check: https://github.com/OCA/account-fiscal-rule/pull/66

luismontalba (approved):

PR #61, opened 2016-12-12
Migrate account-fiscal-rule to V10

yk2kus:

PR #68, opened 2017-10-31
Migrate account_fiscal_position_rule_sale to v10.0

TimLai125:


pedrobaeza:
But this one depends on account_fiscal_position_rule. Where it is?

TimLai125:
@pedrobaeza account_fiscal_position_rule has been migrated and merged in #61, do you mean I need to make any further adjustments on the account_fiscal_position_rule?

pedrobaeza:
Oh, sorry, it was not written in the migration issue.

yostashiro (approved):
Code reviewed. LGTM. 👍

account_fiscal_position_rule_sale/models/sale.py
```diff
@@ -21,15 +22,20 @@ def _fiscal_position_map(self, **kwargs):
...
 
-        kwargs = {
-            'company_id': self.company_id,
-            'partner_id': self.partner_id,
-            'partner_invoice_id': self.partner_invoice_id,
-            'partner_shipping_id': self.partner_shipping_id,
-        }
+        # If the fiscal position of the customer is set, we respect the
+        # setting.
+        super(SaleOrder, self).onchange_partner_shipping_id()
```
JordiBForgeFlow, line 29: I do not understand this part. The notes does not match woth what it then does.  When do you really want to really avoid updating the fiscal position? I would say if the partner_shipping_id has changed, I would apply the new fiscal position.

JordiBForgeFlow (changes requested):

account_fiscal_position_rule_sale/models/sale.py
yostashiro: @jbeficent As I look at the code again, I agree that this part should be improved if we are to migrate this module.

What I just realized is that the standard function actually since V9 covers the particular requirement we were trying to cover using this module - automatically propose the correct fiscal position based on the destination state.  Ref: https://github.com/odoo/odoo/blob/be7927d629b765c4f3eeb3ada11806ba75d909ca/addons/account/models/partner.py#L94-L134

The main function the standard Odoo may be missing is the ability to consider the address of the source location.  Therefore, if we are to cover this missing feature, I guess we should probably review the overall design of the modules in this repo as the large part of the functions these modules are trying to deliver are redundant with what standard Odoo now provides.  I guess just extending the fiscal position model (instead of having fiscal position rule model) would make more sense.

yostashiro (changes requested):
I need to revoke my earlier approval. ^^;

rvalyi:
cc @renatonlima I think you were working on that one too, may be you can check

TimLai125:
According the comments from @yostashiro , most of the functionalities of this module are covered by standard.
I will close this PR for now if anyone is interested please feel free to continue my work.

Issue #69, opened 2017-11-17
[RFC] account_tax_cache: Caching of account.tax lookups


lasley:
This module will provide a method to cache sales tax lookup results at the sale level, for subsequent use during line item lookups.

This will be highly beneficial when connecting to external tax systems in order to reduce API calls. It will also allow for the submission of order level data, such as freight, which can be taxable in some instances. To constrast this, the current Odoo design has lookups being performed at the line item level in most instances.

A preliminary design for this can be seen in my [TaxJar connector PR](https://github.com/LasLabs/odoo-connector-taxjar/pull/1), which is going to be my incubator for this feature as I'm designing it. We'll probably see a few more base modules come out of this as I go along as well.

At the moment, I'm finding it pretty hard to support different record types for the tax caching (eg sales vs purchases). My needs are only for sales tax, and I don't really have an idea of how purchase taxing would even work. Is anyone currently using or thinking of a tax connector for anything other than sales tax?

lasley:
cc @rvalyi

lasley:
Closing to track in #70

PR #70, opened 2017-12-04
[ADD] base_tax_connector: Logic for tax connectors

lasley:
This module provides base logic to allow for easy addition of tax connectors providing external rate lookups. It implements all aspects of the caching mechanism described in #69, as well as some additional features which I will outline below shortly.

This is a WIP as I implement in a tax connector (https://github.com/LasLabs/odoo-connector-taxjar/pull/1) to verify that theory works. So far, so good though!

Currently supported is:
* `sale.order`
* `account.invoice`

Planned on supporting:
- [x] `purchase.order`

An interface/adapter mechanism is provided here, which allows for the easy support of new models. The following is required to support a new model:
* [Inherit `account.tax.rate` and make changes documented in class-level docblock](https://github.com/LasLabs/account-fiscal-rule/blob/c78e1c26b63dc567670ff319b647f3e0e3c01e3d/base_tax_connector/wizards/account_tax_rate.py#L13)
* [Inherit the target model's line item and override the tax computation to prime the cache at the parent level](https://github.com/LasLabs/account-fiscal-rule/blob/c78e1c26b63dc567670ff319b647f3e0e3c01e3d/base_tax_connector/wizards/sale_order_tax_rate.py)
  * This is required because most of the Odoo price/tax computations are performed at the line item level, but tax calculations are relevant for the parent record as a whole. Calculating at the parent level allows us to account for things like freight, tax tiering, etc. It also reduces the amount of API calls that are sent to any external system.
* [Create an adapter model providing the following methods](https://github.com/LasLabs/account-fiscal-rule/blob/c78e1c26b63dc567670ff319b647f3e0e3c01e3d/base_tax_connector/wizards/sale_order_tax_rate.py):
  * `get_rate_lines` - return a list of rate line value dictionaries. It's recommended to create the values themselves in a separate function (`get_rate_line`), but not required.
  * `get_company` - return the company the record belongs to
  * `get_partner`- return the partner the record references
  * `get_untaxed_amount` - return the total price of the record, minus any taxes (the subtotal)

Still need to do:
* Interface for: 
  - [x] New tax transaction
  - [x] ~Update tax transaction~ Skipped for v1; invoices are immutable
  - [x] Refund tax transaction

Still need to verify:
- [X] What to do in event of record being confirmed with one price, then payment being received after a tax change (waiting on external expert's opinion, plus [community comments](https://github.com/OCA/account-fiscal-rule/pull/70#issuecomment-349069000))
  * Nothing
  * > Most states prefer or require that taxes are calculated based on the principles of accrual accounting, which would necessitate that the tax be calculated based on the rate in effect at the time the sale was recognized (not when the cash was received) and reported in the period during which the recognition of the sale occurred. Some states will allow sellers to report on a cash basis in lieu of accrual basis, but usually this option is reserved for smaller-scale vendors and may require special permissions.


base_tax_connector/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+    "category": "Connector",
+    "website": "https://laslabs.com",
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "account",
+        # @TODO: Figure out how to install below in a hook when testing.
+        "purchase",
```
lasley: Break these two into glue modules

base_tax_connector/models/account_tax.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class AccountTax(models.Model):
+
+    _inherit = 'account.tax'
+
+    amount_type = fields.Selection(
+        selection_add=[('cache', 'Cached Data')],
+    )
+    cache_location = fields.Char(
```
lasley: Remove

base_tax_connector/models/account_tax.py
```diff
@@ -0,0 +1,39 @@
...
+        help='This is the location of the cache store. It is an '
+             'arbitrary string that is primarily meant for connectors to '
+             'identify themselves.',
+    )
+    cache_by_group = fields.Boolean(
```
lasley: Remove

base_tax_connector/README.rst
```diff
@@ -0,0 +1,57 @@
+|License AGPL-3| | |Build Status| | |Test Coverage|
```
lasley: LGPL

lasley:
Does anyone in the community currently have a strategy for the handling of tax rate changes between when the relevant record (sale) is confirmed, and when payment is actually received on that record? This seems like kind of a black hole that is never mentioned in any of the Odoo tax docs.

lasley:
Alright so regarding my question of whether tax rate changes between order and payment has been answered by the experts, at least for the US. I welcome any additional insight for other countries.

> Most states prefer or require that taxes are calculated based on the principles of accrual accounting, which would necessitate that the tax be calculated based on the rate in effect at the time the sale was recognized (not when the cash was received) and reported in the period during which the recognition of the sale occurred. Some states will allow sellers to report on a cash basis in lieu of accrual basis, but usually this option is reserved for smaller-scale vendors and may require special permissions.

lasley:
Hmmm so an interesting conundrum I'm running into. Once an invoice has been quoted its taxes, they are stored in the `account.invoice.tax` records as a grouped amount for all line items per specific tax. 

Does anyone know of an existing strategy to maintain the tax computations at the invoice line item level instead? I really feel like this should exist and I'm just missing something.

lasley:
Alright I am done with this in terms of changes. Still a few tests to add, but this is ready to review.

Note that an issue known for launch of this version is the invoice line naivety mentioned in https://github.com/OCA/account-fiscal-rule/pull/70#issuecomment-350033429. This only hurts granularity, and can be added later without any consequence. It's not an easy task though, so I think deferring it makes sense.

rvalyi:
I really try hard to get time to review this a bit... cc @renatonlima

lasley:
Thanks @rvalyi. FYI we've released this as-is in our own repos in order to accommodate the launch of our tax connector. We went live with the connector last night for one customer and have another planned for Friday. Given that, I'd say we have a solid idea of functionality outside of test environments by end of next week.

lasley:
The big question here IMO isn't really functionality, but more the amount of API requests being sent to the server. Using standard Odoo mechanisms, I was at about 40-50 API calls by the time a sale order with 3 lines had a validated invoice. That's why the cache here is so important, but I know there's still some duplication because I can't actually write to the database during onchanges. Not sure how to circumvent that without getting dirty though.

rvalyi:
Hello @lasley and all. I just wanted to tell that yesterday I finally got deep into that and I understood how it works, the way you try to limit API calls by grouping things 1st and caching rate and rate.lines before using them in the tax _compute_all method. This looks smart and complex. So far what I see makes sense, the code looks clean enough...

My goal was to analyse the differences with the fiscal rules modules we created and if there were things we could benefit from for the Brazilian localization. The biggest learning is that we better do like you did: don't create account.tax for every possible % but instead use just several account.tax and make their % dynamic just like you did in the _compute_all method. Now in my opinion the rate and rate.line caches only make sense in an API connector logic (something we don't have or we don't need in Brazil for instance). Basically my only doubt is whether the rate and rate.lines is generic enough or if it is too much tied to the TaxJar API and the USA tax system.

However, with @lasley out of Odoo business now, I wonder if this PR will ever be merged... I cannot say I tested things deep enough to approve it straight at this stage. And without this and Laslabs, I'm also quite worried Odoo may not make its way in a clean and open source way in the USA, or at least suffer quite a lag, specially with e-commerce....

Well at least I let you know that I worked on it, I'm mostly positive about it. If you want me to spend more time on this, you can contact me, or even hire me (via Akretion) to work more on this... For now, I already have much work with our Brazilian tax system in Odoo I'm afraid...

PR #93, opened 2019-05-21
[8.0][ADD] account fiscal position type

legalsylvain:
New module that adds ``type_position_use`` field on ``account.fiscal.position`` model. possible value ('sale' / 'purchase' / 'all' as for ``account.tax``). If set, fiscal position will be available on out or in invoices, limiting incorrect values.

Note : this module includes a template mechanism that will work if https://github.com/OCA/OCB/pull/850 is merged.


account_fiscal_position_type/models/account_fiscal_position_template.py
```diff
@@ -0,0 +1,30 @@
...
+        selection=_TYPE_POSITION_USE_SELECTION, default='all')
+
+    @api.model
+    def generate_fiscal_position(
+            self, chart_temp_id, tax_template_ref, acc_template_ref,
+            company_id):
+        return super(AccountFiscalPositionTemplate, self.with_context(
+            chart_template_id=chart_temp_id)).generate_fiscal_position(
+            chart_temp_id, tax_template_ref, acc_template_ref,
+            company_id)
```
pedrobaeza, line 30: Why not simply rewrite after the call to super the type?

legalsylvain: Hi @pedrobaeza,

Because ``generate_fiscal_position`` is an @api.model function and doesn't return just created ``account.fiscal.position``. ([Ref](https://github.com/odoo/odoo/blob/8.0/addons/account/account.py#L2962)) So I can not know which ``account.fiscal.position`` I have to update. 

Maybe I missed something, but for me, passing the context in the ``create`` function was the only possibility to do it. If you have an idea, feel free to propose !

About that module : 
- There is no bug without the patch (proposed on OCB). Just the fiscal.template value is ignored. It's a limitation, noted in the roadmap section. 
- When I'll port this module into 10 / 12 serie, I'll can have the possibility to handle without patch this feature, as context is allways passed in new API. 

so my proposal for OCA is to accept this PR as it. (I set ``'development_status': 'Beta'`` for that purpose). And improve it when porting. (well of course, only if people think it is an interesting feature ;-))

In all case, thanks for your review !

pedrobaeza: Another option, although obviously is not the most elegant, but works in this tricky cases:

- Search for all existing fiscal positions before calling super.
- Call super.
- Search again for all existing fiscal positions now. The difference will get you the created fiscal position.

I gave a talk in Spanish Jornadas de Odoo with tips like this one. You can see it at https://odoospain.odoo.com/event/jornadas-odoo-2016-2016-06-16-2016-06-17-4/track/como-realizar-un-modulo-de-odoo-compatible-con-todo-152

Maybe I can repeat a similar talk in Spanish OCA Days, but I think you aren't coming. Aren't you?

legalsylvain: Hi. I understand your PoV. And I think what you propose is a right way to do it in v8. However, I have no time to change this module especially since this code is the correct one, when I will migrate this module for v12.
I so close this PR. Feel free to reopen it, if you think that having this module in V8 (in this state), is better than nothing.

PR #97, opened 2019-08-16
Update Postgres configuration on Travis to avoid error.

mbcosta:
Update Postgres configuration on Travis to avoid error.

cc @renatonlima @rvalyi

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "2.7"
 
 addons:
+  postgresql: "9.6"
   apt:
     packages:
       - expect-dev  # provides unbuffer utility
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,13 +17,10 @@ addons:
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
-  - TRANSIFEX_USER='transbot@odoo-community.org'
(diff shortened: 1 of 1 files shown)


pedrobaeza:
You can take the occasion to update whole file with one less build and removing old Transifex stuff

pedrobaeza:
Why putting 9.2 instead of 9.6?

mbcosta:
Hi @pedrobaeza , I receive the same error with 9.6, I just going to check if it's a version problem ( looking in other repo projects in OCA at version 10 ) or do you think it's is not relation ?

danimaribeiro:
You can use:
```
services:
  - postgresql
```

rvalyi:
@mbcosta you made a mistake, look twice at the master template:
https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml#L13
you put postgres inside the ```apt``` section but it's inside the ```addons``` parent section instead (it's a service, not an apt package here).

.travis.yml
```diff
@@ -16,9 +17,7 @@ addons:
 
 env:
   global:
-  - VERSION="10.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
-  - TRANSIFEX_USER='transbot@odoo-community.org'
-  - secure: "FmJknQWKDSXzrQZm5+PNNhQwEJzA4IVQpT6UASSthvQA3i88nsX3khsHBRAoQhJr7NTinIp4nIvsvVvsrKXSnbDNY0bScrx8dCDD2weKZnsrD4Khh6Gn3pFwCrwJiPccnMUnFuBxbJzV/tZRrupQIcIfFg8AApCYJ4Qm+nb5XuA="
+  - VERSION="10.0" TESTS="0" LINT_CHECK="0"
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
```
pedrobaeza: Remove this one and add `MAKEPOT="1"` in the last build.

mbcosta:
Thanks all for help

PR #106, opened 2020-03-10
Fiscal classification readme

EBII:
Update index.html 
update readme/*rst file with img path

restore initial Readme template


dreispt (changes requested):
The changes files are automatically generated from the fragments in `/readme`, so there is little point in these changes.

legalsylvain:
hi @EBII. Could you reduce your PR, removing changes in readme.rst and index.html file ? 

thanks !

dreispt:
@legalsylvain Actually, I see now a lot of PRs with the same pattern, because precommit hooks automatically generate the README.
So I must say we shouldn't mind about that.

dreispt (approved):

legalsylvain:
Hi @dreispt. Are you sure ? I don't see it in the pre-commit file of repo I checked ?

Anyway, I see it as a problem if precommit update such files, because it will generate a lot of conflict. (like here, that should be solved, @EBII )

dreispt:
@EBII To update the README, please use the fragments in the `/readme` subdirectory.
You're welcome to propose a new PR.

PR #112, opened 2020-04-24
12.0 FIX prd_classif: set taxes on all companies sharing classif

bealdav:
Missing sudo prevent to see all taxes in other companies and lead to write bad values in product tax fields

Bug appears only when no parent_id in your companies

cc @legalsylvain @PierrickBrun

### DIFF
--- a/account_product_fiscal_classification/demo/account_product_fiscal_classification.xml
+++ b/account_product_fiscal_classification/demo/account_product_fiscal_classification.xml
@@ -27,6 +27,18 @@
         <field name="purchase_tax_ids" eval="[(6, 0, [ref('account_tax_purchase_1')])]"/>
     </record>
     <record model="account.product.fiscal.classification" id="global_fiscal_classification_1">
+        <field name="name">Demo Global Fiscal Classification 1</field>
+        <field name="company_id" eval="0"/>
+        <field name="purchase_tax_ids" eval="[(6, 0, [ref('account_tax_purchase_1'), ref('purchase_tax_44_other')])]"/>
+        <field name="sale_tax_ids" eval="[(6, 0, [ref('account_tax_sale_1'), ref('sale_tax_44_other')])]"/>
+    </record>
+    <record model="account.product.fiscal.classification" id="global_fiscal_classification_2">
+        <field name="name">Demo Global Fiscal Classification 2</field>
+        <field name="company_id" eval="0"/>
+        <field name="purchase_tax_ids" eval="[(6, 0, [ref('account_tax_purchase_1'), ref('purchase_tax_7_other')])]"/>
+        <field name="sale_tax_ids" eval="[(6, 0, [ref('account_tax_sale_2'), ref('sale_tax_7_other')])]"/>
+    </record>
+    <record model="account.product.fiscal.classification" id="global_fiscal_classification_1">
         <field name="company_id" eval="False"/>
         <field name="name">Demo Global Fiscal Classification 1</field>
     </record>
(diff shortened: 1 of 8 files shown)


account_product_fiscal_classification/demo/account_product_fiscal_classification.xml
```diff
@@ -26,6 +26,18 @@
         <field name="company_id" ref="base.main_company"/>
         <field name="purchase_tax_ids" eval="[(6, 0, [ref('account_tax_purchase_1')])]"/>
     </record>
+    <record model="account.product.fiscal.classification" id="multicpny_fiscal_classif_1">
```
legalsylvain: Could you harmonize with existing demo data ? record id, record name, etc...

```
    <record model="account.product.fiscal.classification" id="global_fiscal_classification_1">
        <field name="company_id" eval="False"/>
        <field name="name">Demo Global Fiscal Classification 1</field>
    </record>
```

account_product_fiscal_classification/models/product_template.py
```diff
@@ -140,3 +145,18 @@ def _check_access_fiscal_classification(self, vals):
                     "You can not use the fiscal classification '%s' because"
                     " you're not member of the group '%s'.") % (
                         classification.name, group.name))
+
+
+def tax_infos(product):
```
legalsylvain: could you remove this function as it seems to exist only for debug purpose ?

legalsylvain (commented):
some remarks. Make sense otherwise. Thanks !

account_product_fiscal_classification/models/product_template.py
bealdav: if I remove I no more can use in logger.debug(), isn't it ? I put in test, ok for that

account_product_fiscal_classification/demo/account_product_fiscal_classification.xml
bealdav: Thanks for the review.
Here about names, I don't know what harmonize could mean ? multicpny_fiscal_classif_1 vs global_fiscal_classification_1 
What is the added value ? the more significant ? but if you want, I can this time.
If you want more string replaced please, could you use github suggestions. I don't know your taste in strings naming ;-) Thanks

account_product_fiscal_classification_test/tests/test_multicompany.py
```diff
@@ -0,0 +1,65 @@
...
+                    "cpny_onlyshare_classification"
+                ).id
+            }
+        )
+        # check with same data
+        check_sale_purchase_tax(classif2, write_classif=False)
+        check_sale_purchase_tax(classif1)
+
+
+def tax_infos(product):
```
legalsylvain: Hi. Well, i don't understand why this function is usefull. I mean, making log in test file. test file is only for assert things. don't you think ?

account_product_fiscal_classification/demo/account_product_fiscal_classification.xml
legalsylvain: just having the same kind of name for a single model. But you did it. Thanks.

account_product_fiscal_classification/models/product_template.py
legalsylvain: But what is the reason to add a 20 lines function that makes a log ?

account_product_fiscal_classification_test/tests/test_multicompany.py
bealdav: Except this point, others are ok ?

legalsylvain: yes !

bealdav: ok I will remove

legalsylvain (approved):
Hi. Thanks for this great patch. LGTM. (I permit me to remove some useless diff). You can squash all the commits together.

thanks !

sebastienbeau (approved):
Thanks for the fix. LGTM (code review)

PR #117, opened 2020-07-16
[FIX] account_avatax: correct Avatax amount lookup when using line discount

dreispt:
Making further tests, I found that the tax computation was not correct when using line discounts.
The image below shows an example Invoice reproducing the error.
This changes fixes the issue.

image

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -289,6 +289,34 @@ class AccountMoveLine(models.Model):
 
     avatax_amt_line = fields.Float(string="AvaTax Line", copy=False)
 
+    def _get_avatax_price_unit(self):
+        """
+        Return the company currey unit price, after discounts,
+        to use for Tax calculation.
+
+        Code extracted from account/models/account_move.py,
+        from the compute_base_line_taxes() nested function.
+        """
+        self.ensure_one()
+        base_line = self
+        move = base_line.move_id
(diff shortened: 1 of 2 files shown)


max3903 (approved):

b-kannan (commented):
@dreispt 

Why use sign in the function and reverse the return value if we are going to use the sign in the caller function _avatax_prepare_line?

if the _get_avatax_price_unit() function returns the value with correct sign, wouldn't that be sufficient to determine the sign for the avatax_amount in compute_all function.  Is this code specific for refund processing?

dreispt:
@b-kannan The Invoice Lines now are also an Accounting Moves, and there they carry the inverted signs.
The `_get_avatax_price_unit()` works from the "Invoice" perspective, so it return positive amount for Invoices, and so needs to invert the account move signal.

The `compute_all()` is used by the accounting logic, and receives amount with the Account Move perspective.
So it needs to invert them before comparing with "invoice lines".

Finally, the Tax amount must be added or subtracted depending if this is an invoice or a Credit Note.
I used copysign here as a simple way to do the right calculation, without adding more complex checks.

To ocnfirm this we should Post an invoice and then create a Credit Note on it, and chck on the accounting moves that the Tax amount sign (DR/CR) is the expected one.

b-kannan:
@dreispt : The comments make sense. Accepted.

PR #119, opened 2020-08-19
[IMP] account_avatax: address validation uses active company, not the Partner's company

dreispt:
The previous behavior caused confusion in multi-company environments, and was change per user request.
Forward port of a v12 change.


b-kannan (approved):
Good with this change.

agyamuta (approved):
LGTM!. Thanks.

PR #120, opened 2020-08-19
[FIX] account_avatax: Credit Notes were being sent as positive amounts to Avatax service

dreispt:
The "sign" is supposed to be +1 for outgoing docs and -1 for incoming docs.
An "out_refund" is still an +1 outgoing doc, like "out_invoice".

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -128,7 +128,7 @@ class AccountMove(models.Model):
         Prepare the lines to use for Avatax computation.
         Returns a list of dicts
         """
-        sign = self.type == "out_invoice" and 1 or -1
+        sign = 1 if self.type.startswith("out") else -1
         lines = [
             line._avatax_prepare_line(sign, doc_type)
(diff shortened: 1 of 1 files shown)


b-kannan (changes requested):
@dreispt : The self.type refers to the journal entry or invoice?

It is account.invoice that has the type field and it can be one of 4 types. Customer invoices and credit notes, vendor bills and vendor refunds. Vendor bills and customer credit notes are outgoing (-) and the other two are incoming (+).

Is this part only for customer focused transactions? Otherwise, the check needs to be different.

dreispt:
@b-kannan This is v13, so it is both: a Journal Entry that also represents an Invoice.
Avatax only supports "outbound" customer documents now, so this check doesn't do much.
But in the future we might need to support "incoming" vendor documents, for Use Tax purposes, so I decided to leve it in the code.

b-kannan (approved):
@dreispt I am good with this.

PR #121, opened 2020-08-20
[IMP] Use exemption codes set at customer address level 

dreispt:
Forward port of 12.0 improvement.
Also includes forward port of fix to Exemptions not applied if a tax rate had been previously computed.


bodedra (approved):
Code review :+1:

dreispt:
May be superseded by #126

b-kannan (commented):
I am confused. Isn't the exemption_code_id on partner record supposed to be property field? I thought the field name was property_exemption_code_id in an earlier commit.

dreispt:
Yes, this was refactored by #126 and is obsolete. Closing.

PR #122, opened 2020-08-20
[FIX] account_avatax: Don't copy customer code and address validatin, when duplicating a Partner

dreispt:


dreispt:
Fixed the merge conflict, should be good to go.

atchuthan (approved):
:+1:

PR #123, opened 2020-08-20
[FIX] account_avatax: use for destination address the ship-to comped field, honoring the Tax on Shippling Address flag

dreispt:
The destination address should be taken from the computed field for that, not directly from the shiping-to address.
This was correct in v12, and the error introduced in the v13 port.

The expected behavior is, when the "Tax on Shipping Address" is unchecked, then the destination address should be the Customer one and not the Delivery Address.

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -151,7 +151,7 @@ class AccountMove(models.Model):
             doc_type,
             self.partner_id,
             self.warehouse_id.partner_id or self.company_id.partner_id,
-            self.partner_shipping_id or self.partner_id,
+            self.shipping_add_id or self.partner_id,
             taxable_lines,
             self.user_id,
             self.exemption_code or None,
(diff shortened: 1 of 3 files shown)


account_avatax/models/account_move.py
```diff
@@ -151,7 +151,7 @@ def _avatax_compute_tax(self, commit=False):
             doc_type,
             self.partner_id,
             self.warehouse_id.partner_id or self.company_id.partner_id,
-            self.partner_shipping_id or self.partner_id,
+            self.shipping_add_id or self.partner_id,
```
atchuthan: There is no shipping_add_id in account.move object, please change it to tax_address_id

atchuthan (changes requested):

account_avatax/models/account_move.py
atchuthan: Sorry, I got confused with the tax_address_id definition in sale.order

@dreispt We could follow the same naming convention across account_avatax and account_avatax_sale

atchuthan (approved):

account_avatax/models/account_move.py
dreispt: Bad names are bad code. Done as suggested

atchuthan (approved):
:+1:

PR #124, opened 2020-08-20
[FIX] account_avatax: v13 new syntax for company record rules

dreispt:


dreispt:
Company record rule syntax changed in v13.
Example: https://github.com/odoo/odoo/blob/13.0/addons/crm/security/crm_security.xml#L73

atchuthan (approved):

PR #125, opened 2020-08-25
[FIX] account_avatax: Ensure Avatax not called if no line has Avataxes

dreispt:


dreispt:
The issue here is that the `lines` list, to send to Avatax for tax computation, could contain empty elements.
A list containing an empty element is not an empty list, and an Avatax could be triggered to Avatax with no data to process, causing an error.

atchuthan (approved):

PR #131, opened 2020-09-22
[FIX] account_avatax: 100% discount lines should also be sent to Avatax

dreispt:
A Sale Order with 100% on all lines would not compute taxes, since no line would be sent to Avatax.
Ww should send all SO lines to Avatax, even if total amount is zero.


atchuthan (approved):

b-kannan (approved):
@dreispt 

Q: Is there a reason why all lines must be sent to Avalara API? 
A: There are certain situations where the tax authorities expect taxes to be computed on the non-discounted amount and collected. So, in cases where there is 100% discount, the line may still be taxable.

PR #135, opened 2020-10-02
[FIX] account_avatax: Regular users can't set exemption status correctly (onchange does not work on readonly fields)

dreispt:


dreispt:
Note: this is using the new "computed writable fields" feature, introduced in v12.
I'm calling them `_compute_onchange`.
These are slated to replace onchange in the future.

b-kannan (approved):

PR #137, opened 2020-10-02
[IMP] account_avatax: official Avalara certification awarded 🍾

dreispt:


b-kannan (approved):
@dreispt : Thank you and congratulations.

PR #141, opened 2020-10-21
12.0 IMP prd_fisc_classif: track change on product model

bealdav:
cc @legalsylvain @hparfr @florian-dacosta

### DIFF
--- a/account_product_fiscal_classification/models/product_template.py
+++ b/account_product_fiscal_classification/models/product_template.py
@@ -17,6 +17,7 @@ class ProductTemplate(models.Model):
     fiscal_classification_id = fields.Many2one(
         comodel_name='account.product.fiscal.classification',
         string='Fiscal Classification',
+        track_visibility="onchange",
         help="Specify the combination of taxes for this product."
         " This field is required. If you dont find the correct Fiscal"
         " Classification, Please create a new one or ask to your account"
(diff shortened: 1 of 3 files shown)


dreispt (approved):

PierrickBrun (approved):

legalsylvain (changes requested):
I dont think it should be done in main module but in custom one. AFAIK, taxes changes are not tracked in Odoo core. I dont see why this field shoud be tracked and not all the others.
Don-t you think ?

bealdav:
Thanks for the review

> I dont see why this field shoud be tracked and not all the others.

Because it have a direct impact on a legal feature in case of sharing catalog in multi company: tax in invoicing and it's very critical to track changes on it. But if on your use case it's useless, please close the PR, i'll add a custom for it. Thanks

PierrickBrun:
> I dont think it should be done in main module but in custom one. AFAIK, taxes changes are not tracked in Odoo core. I dont see why this field shoud be tracked and not all the others.
> Don-t you think ?

That's a good point but IMO this module exists to mitigate tax errors, so this should be a nice addition to have the tracking.

legalsylvain (approved):
Make sense. Thanks for the explanation.

florian-dacosta (approved):

hparfr (approved):

mourad-ehm (approved):

rvalyi (approved):

legalsylvain:
7 approvals ! it's a plebiscite

PR #143, opened 2020-10-26
[FIX] account_fiscal_position_type : do not crash if domain is not set

legalsylvain:
minor fixes. for the time being, when we click on supplier / refund / create an error is raised, due to the lake of domain in the returned value. 

this PR fixes that bug.

CC : @quentinDupont #GRAPOCA (ligne 192)


dreispt (approved):

quentinDupont (approved):

astirpe (approved):

legalsylvain:
@dreispt (or other PSC) could you merge ?

rvalyi (approved):

legalsylvain:
thanks !

PR #145, opened 2020-10-29
[FIX] account_avatax_sale: Field 'is_add_validate' definition not available

SodexisTeam:
Either we should remove the "is_add_validate" field from _onchange_partner_shipping_id or add this field to account.move


dreispt:
Is that field useful?

atchuthan:
The field is not used in any Avatax API.  Also, this field was hidden in the older version.

dreispt:
If if is not useful, maybe we just just remove it completely.

dreispt (approved):

PR #146, opened 2020-11-02
[FIX] account_avatax: Update taxes only for Draft Invoices

atchuthan:
Minor fix: Only update taxes with Avatax API for Draft invoices.

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -212,7 +212,9 @@ class AccountMove(models.Model):
         Forces computation of the Invoice taxes
         """
         for invoice in self:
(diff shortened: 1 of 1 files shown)


dreispt (approved):
This looks good, and it could go further: the Server Action should go away, and the filter should be in the ORM code.

atchuthan:
I got your point, I made changes such that either condition of state == 'draft' or commit==True if available.  Because commit=True will be sent after the Invoice is validated.

dreispt:
Thank you, this is an improvement, and achieves the goal of preventing of accident.
To also protect against ill intended RPC calls a small refactoring would be needed, but I guess that is out of scope for this fix.

PR #150, opened 2020-11-30
[MIG] account_avatax

dnplkndll:


dnplkndll:
@dreispt the exemptions work needs split up to remove the sign dependancy. It looks like I got too much history in this mig branch. Any advice is always appreciated. I did load the avalara.salestax and product.tax.code but only tested the ping so far. As I get some test automation setup we could probably come up with the use cases on the agricultural item types withe ag exemption mapping in state/counties.

account_avatax/models/avalara_salestax.py
```diff
@@ -136,6 +136,10 @@ def _get_avatax_supported_countries(self):
         help="Allows ean13 to be reported in place of Item Reference"
         " as upc identifier.",
     )
+    avatax_company_id = fields.Char(
+        'Company ID',
+        help="The company ID as defined in the Admin Console of AvaTax",
+    )
```
dreispt, line 146: This info is already there, as `company_code `

account_avatax/__manifest__.py
```diff
@@ -1,7 +1,8 @@
 {
-    "name": "Avalara Avatax Connector",
+    "name": "Taxes using Avalara Avatax API",
     "version": "14.0.1.0.0",
-    "author": "Fabrice Henrion, Odoo Community Association (OCA)",
+    "author": "Open Source Integrators, Fabrice Henrion, Odoo SA,"
```
dreispt: Please keep Fabrice as the author.
There is some back and forth email history about this.

dreispt: I mean, remove OSI and Odoo from the authors. Thanks.

account_avatax/models/account_move.py
```diff
@@ -138,10 +138,8 @@ def _avatax_prepare_lines(self, doc_type=None):
     def _avatax_compute_tax(self, commit=False):
         """ Contact REST API and recompute taxes for a Sale Order """
         self and self.ensure_one()
+        Tax = self.env["account.tax"]
         avatax_config = self.company_id.get_avatax_config_company()
-        if not avatax_config:
-            # Skip Avatax computation if no configuration is found
-            return
```
dreispt: You'll get errors when running Odoo tests if you remove this.

dreispt:
@dnplkndll Thank you for working on this.
Yes, adding tests it pretty much needed and appreciated.
I've been thinking on ways to make this module smaller, and I'm considering moving the "force tax amount" feature into its own module. It could also be useful for other tax conenctors.
And we will need to run this through Avalara certification before being able to display the certified logos in the README.

Maybe we could merge this as "alpha" state, while woring on the above points.
If so, we should not display the certification badges in the README until we go through the certification process.

account_avatax/models/avalara_salestax.py
dnplkndll: avatax_company_id is their reference, used in api calls and useful to have for the ecom integrations. company_code is our reference value

image

account_avatax/models/account_move.py
dnplkndll: ah will re-add, I have that in my migration.yml

dreispt:
#160 has a clean history extraction, and is ready to have the migration changes added on top of it.
You're welcome to rebase on it, or create a new PR based on #160 and then copy over you changes.

atchuthan:
@dreispt we rebased the commit based on https://github.com/OCA/account-fiscal-rule/pull/160 and added our migration fixes in this PR itself

cybernexus:
where is this pull request at in terms of being merged?

dreispt (approved):
There are a few bugs to fix, but this is deployed in several instances and is working for the most part.
Bug fixing will work best as separate PRs.
So let's merge it!

PR #152, opened 2020-12-02
[MIG] 13.0 account fiscal classification

PierrickBrun:


PierrickBrun:
I squashed commits but there is a lot of history in this module

PierrickBrun:
> Too much honour! BTW: I think the OCA likes their hooks in hooks.py https://github.com/OCA/maintainer-tools/blob/master/template/module/hooks.py

Indeed, this is better like this

dreispt (approved):

PR #153, opened 2020-12-03
[FIX] account_avatax: minor Fix

SodexisTeam:
'result' variable undefined

### DIFF
--- a/account_avatax/models/avalara_salestax.py
+++ b/account_avatax/models/avalara_salestax.py
@@ -279,6 +279,7 @@ class AvalaraSalestax(models.Model):
 
     def commit_transaction(self, doc_code, doc_type):
         self.ensure_one()
+        result = False
         if not self.disable_tax_reporting:
             avatax = self.get_avatax_rest_service()
             result = avatax.call(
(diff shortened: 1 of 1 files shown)


account_avatax/models/account_move.py
```diff
@@ -142,6 +142,9 @@ def _avatax_compute_tax(self, commit=False):
         if not avatax_config:
             # Skip Avatax computation if no configuration is found
             return
+        if not avatax_config.disable_tax_reporting:
+            # Skip Avatax Tax calculation if Tax reporting is disabled
+            return
```
dreispt: The `disable_tax_reporting`is supposed to not commit transaction, but should still perform the API calls.
So this is not functionally correct.
Am I seeing this right?

atchuthan: Sorry, I meant to disable the check when disable_tax_calculation is disabled.

atchuthan: Yes, removed this check as it is already checked in get_avatax_config_company

dreispt (approved):

PR #154, opened 2020-12-03
[MIG] 14.0 account_product_fiscal_classification

PierrickBrun:
Includes https://github.com/OCA/account-fiscal-rule/pull/152


legalsylvain (approved):
H @PierrickBrun. Thansk for porting this module. 
I checked the diff. (quite hard to see all the change, with all the black, isort, etc...)

I saw no changes, regarding the V12. (except the classical adaptation to V14 framework)
right ?

PierrickBrun:
> H @PierrickBrun. Thansk for porting this module.
> I checked the diff. (quite hard to see all the change, with all the black, isort, etc...)
> 
> I saw no changes, regarding the V12. (except the classical adaptation to V14 framework)
> right ?

You are right, This is only classical adaptations to the framework.
The only change I did beside that is modifying the "active" field, using the ribbon to be more consistent with Odoo, following this https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

PR #164, opened 2021-02-08
[FIX] account_avatax: fix erros

atchuthan:
1. Do not raise warning on Address verification to avatax, if no configuration is setup in Odoo
2. Do not raise an error when logging is False at Avatax API

@dreispt

### DIFF
--- a/account_avatax/models/avalara_salestax.py
+++ b/account_avatax/models/avalara_salestax.py
@@ -165,6 +165,7 @@ class AvalaraSalestax(models.Model):
             self.service_url,
             self.request_timeout,
             self.logging,
+            config=self,
         )
 
     def create_transaction(
--- a/account_avatax/models/partner.py
+++ b/account_avatax/models/partner.py
@@ -155,6 +155,8 @@ class ResPartner(models.Model):
             )
             return False
         avatax_config = self.env.company.get_avatax_config_company()
+        if not avatax_config:
+            return False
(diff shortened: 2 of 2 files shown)


account_avatax/models/avalara_salestax.py
```diff
@@ -165,6 +165,7 @@ def get_avatax_rest_service(self):
             self.service_url,
             self.request_timeout,
             self.logging,
+            config=self,
```
dreispt, line 168: I assume this is added to avoid error on a call somewhere.
The proper way to go is to fix the call, and not the function signature here.
Thoughts?

atchuthan: https://github.com/OCA/account-fiscal-rule/blob/13.0/account_avatax/models/avalara_salestax.py#L309
https://github.com/OCA/account-fiscal-rule/blob/13.0/account_avatax/models/partner.py#L158

We could have the same parameters in this method too i.e. config=self

atchuthan: We don't need to specify the avatax info as it is obtained from the config file in https://github.com/OCA/account-fiscal-rule/blob/13.0/account_avatax/models/avatax_rest_api.py#L20

dreispt: OK

dreispt (approved):

PR #175, opened 2021-04-09
[IMP] account_avatax: find and apply exemptions for the shipping Country+State

dreispt:


atchuthan (approved):

dreispt:
Thank you @atchuthan. I need to fix Travis before this can be merged.

account_avatax/models/account_move.py
```diff
@@ -13,10 +13,45 @@ class AccountMove(models.Model):
...
+        - ACME Invoicing address, USA CA, no exemption status
+        - ACME Delivery address, USA CA, no exemption status
+
+        Invoice to ACME Invoicing, Shipped to ACME Delivery with be exempt.
+
+        For this to work properly, the "exemption_lock" is no longer supported.
+        """
+        for invoice in self:
+            invoice_partner = invoice.partner_id.commercial_partner_id
+            ship_to_address = invoice.shipping_add_id
```
atchuthan: There is no shipping_add_id in Odoo base code, it is partner_shipping_id

atchuthan (changes requested):

account_avatax/models/account_move.py
dreispt: Fixed, and build is green now

bodedra (approved):
Functional review. :+1: 

Tested with following two ways:

1. Manual invoice
2. Created SO and create invoice.

PR #176, opened 2021-04-09
[14.0][IMP] account_avatax: find and apply exemptions for the shipping Country+State

dreispt:


account_avatax/views/account_move_view.xml
```diff
@@ -27,12 +27,6 @@
                     groups="stock.group_stock_multi_warehouses"
                 />
                 <field name="tax_address_id" />
-                <field name="exemption_locked" />
-                <field name="calculate_tax_on_save" invisible="1" />
-                <field
-                    name="invoice_doc_no"
-                    attrs="{'invisible': [('move_type','!=','out_refund')]}"
-                />
```
atchuthan: Why do we need to delete these fields from view?
we only plan to remove the exemption_locked field, right?

atchuthan (changes requested):

account_avatax/views/account_move_view.xml
dreispt: Right. That was the result from the v13 commit cherry-pick.
I fixed it.

atchuthan (approved):

PR #179, opened 2021-04-12
[FIX] account_avatax: re-CONFIRM of an invoice did not update the Avatax invoice amounts

### LINKED ISSUE #177: [14.0] Avatax: re-Validate invoice with changed amount
Steps:
- Navigate to a posted invoice
- "Set to Draft" -> The invoice is "voided" in Avatax
- Change invoice amount
- "Validate" -> in Avatax the invoice is "commited", but the invoice amount is not updated

dreispt:
Closes #177

Re-Confirming an existing invoice just set the state on Avatax from Void to Committed, without updating the document amounts.
This fixes, by using the "CreateorAdjustTransaction" endpoint instead of "CreateTransaction".

### DIFF
--- a/account_avatax/models/avatax_rest_api.py
+++ b/account_avatax/models/avatax_rest_api.py
@@ -244,15 +244,8 @@ class AvaTaxRESTService:
 
         if doc_date and type(doc_date) != str:
             doc_date = fields.Date.to_string(doc_date)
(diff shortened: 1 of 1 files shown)


b-kannan (approved):
Update to transaction is happening right and verified on the portal transactions.

PR #182, opened 2021-04-16
[14.0][IMP] account_avatax: set Avalara assigned CaseSafeID for this integration

dreispt:


b-kannan (approved):
Approved as the information changed matches the request from Avalara.

PR #186, opened 2021-04-16
[FIX] account_avatax_sale: tax compute before confirm a sale order

### LINKED ISSUE #185: [13.0] account_avatax_sale: tax compute must be happend before confirm a sale order
Steps to reproduce:

- Create a sale order,
- Confirm it

It throws following error:

Screenshot from 2021-04-16 08-25-30

FIX:
Odoo will not allow to change tax after confirm sale order. So super() method must be called at the end of avalara tax computation.

Odoo ref: https://github.com/odoo/odoo/blob/cd9c071c9357cef14635ef094a9f14fc5431956c/addons/sale/models/sale.py#L1810

bodedra:
Closes #185

### DIFF
--- a/account_avatax_sale/models/sale_order.py
+++ b/account_avatax_sale/models/sale_order.py
@@ -191,9 +191,9 @@ class SaleOrder(models.Model):
                     # The Confirm action will be interrupted
                     # if the address is not validated
                     return addr.button_avatax_validate_address()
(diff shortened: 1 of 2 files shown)


bodedra:
@dreispt @atchuthan @max3903

dreispt (approved):

PR #187, opened 2021-04-19
[13.0][FIX] Partner Exemption Code configuration

bodedra:
Steps to reproduce:

- Change name on existing exemption code
- Go to apps and upgrade "account_avatax" module
- It will wipe out changes that we made previously.

This PR will fix it.

### DIFF
--- a/account_avatax/data/avalara_salestax_exemptions.xml
+++ b/account_avatax/data/avalara_salestax_exemptions.xml
@@ -1,4 +1,4 @@
-<data>
+<data noupdate="1">
     <!--
    Partner Exemption Code
    -->
--- a/account_avatax/readme/CONTRIBUTORS.rst
+++ b/account_avatax/readme/CONTRIBUTORS.rst
@@ -5,6 +5,7 @@
 * Open Source Integrators (https://opensourceintegrators.com)
 
   * Daniel Reis <dreis@opensourceintegrators.com>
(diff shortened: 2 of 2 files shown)


bodedra:
@dreispt @atchuthan

atchuthan:
@bodedra 
we are using the code field of exemption.code model as EntityUseCode for the transaction, so I don't think we need to change these configuration data with noupdate="1".
If some customer changed the code wrongly, then the info sent to Avatax transaction will be wrong.

dreispt:
@bodedra I feel @atchuthan has a point.
These are Avatax provided codes, whet is the use case to change them?

bodedra:
@atchuthan @dreispt Code is fine. But in my case, they have different name for same code.

atchuthan:
@bodedra 
If the name you are going to change is a standard name, we can change the name of the exemption.code itself in the data XML file, right?

dreispt:
@atchuthan We should not edit the module's data file directly. I'm OK with adding the noupdate. Edits there are at the customer risk, but so is tax configuration in general.

dreispt (approved):

bodedra:
> @bodedra
> If the name you are going to change is a standard name, we can change the name of the exemption.code itself in the data XML file, right?

@atchuthan We are using submodule structure for OCA module deployment so we never change anything directly.

atchuthan (approved):

marcelsavegnago (approved):

PR #190, opened 2021-04-28
[FIX] account_avatax, account_avatax_sale: Avatax calculation for Line with notes/section

atchuthan:
The index used in enumerate has the wrong list, so the tax is updated in the section/note available in that index.

cc: @dreispt @bodedra @b-kannan @SodexisTeam @stephankeller

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -173,7 +173,7 @@ class AccountMove(models.Model):
         sign = 1 if self.move_type.startswith("out") else -1
         lines = [
             line._avatax_prepare_line(sign, doc_type)
-            for line in self.invoice_line_ids
+            for line in self.invoice_line_ids.filtered(lambda l: not l.display_type)
             if line.price_subtotal or line.quantity
         ]
         return [x for x in lines if x]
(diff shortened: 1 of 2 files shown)


b-kannan (changes requested):
@atchuthan 

The fix in account_move.py doesn't make sens. The first fix (line 176) is having a if statement to it. Since section lines don't have price_subtotal or quantity the section lines should not be picked up. Is this not the case?

The second fix (line 232 of account_move.py) should exclude section lines in the invoice but has been removed.

Not sure what the intent is.

atchuthan:
@b-kannan 
For the 176 line, we remove the section/notes lines that are sent to Avatax.
For the 232 line, we will get the correct tax rate from tax_result_line = tax_result_lines.get(line.id) at line 234 and we use the index to set to the correct line.

i.e. we will get tax_result_line for section/note as None as line.id is not available in tax_result_lines.  So, we will not set the value at the line.

b-kannan:
@atchuthan Thank you.

b-kannan (approved):

dreispt (approved):

PR #171, opened 2021-04-07
[FIX] account_avatax: Full Refunds were not commited to the Avatax service

dreispt:
Closes #170 
The full refund logic calls `_post()`, not `action_post`, so the Avatax logic must move there.

Closes #167
Only apply Avatax logic for sales invoices/refunds.
This is extension friendly, to allow future support for vendor bills or other document types.

PR #191, opened 2021-05-01
[FIX] account_avatax: full refund, closes #171

dnplkndll:
iterate to correct posting in batched, e.g. bank statement posting

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -287,24 +287,26 @@ class AccountMove(models.Model):
         return self.is_sale_document()
 
     def _post(self, soft=True):
-        if self.is_avatax_calculated():
-            avatax_config = self.company_id.get_avatax_config_company()
-            if avatax_config and avatax_config.force_address_validation:
-                for addr in [self.partner_id, self.partner_shipping_id]:
-                    if not addr.date_validation:
-                        # The Validate action will be interrupted
-                        # if the address is not validated
-                        return addr.button_avatax_validate_address()
-            # We should compute taxes before validating the invoice
-            # to ensure correct account moves
-            # However, we can't save the invoice because it wasn't assigned a
-            # number yet
-            self.avatax_compute_taxes(commit=False)
+        for invoice in self:
+            if invoice.is_avatax_calculated():
+                avatax_config = self.company_id.get_avatax_config_company()
+                if avatax_config and avatax_config.force_address_validation:
+                    for addr in [self.partner_id, self.partner_shipping_id]:
+                        if not addr.date_validation:
+                            # The Validate action will be interrupted
+                            # if the address is not validated
+                            return addr.button_avatax_validate_address()
+                # We should compute taxes before validating the invoice
+                # to ensure correct account moves
+                # However, we can't save the invoice because it wasn't assigned a
+                # number yet
+                invoice.avatax_compute_taxes(commit=False)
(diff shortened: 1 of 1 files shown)


dnplkndll:
Review requested
@dreispt @atchuthan @SodexisTeam

account_avatax/models/account_move.py
```diff
@@ -287,24 +287,26 @@ def is_avatax_calculated(self):
...
+                        if not addr.date_validation:
+                            # The Validate action will be interrupted
+                            # if the address is not validated
+                            return addr.button_avatax_validate_address()
+                # We should compute taxes before validating the invoice
+                # to ensure correct account moves
+                # However, we can't save the invoice because it wasn't assigned a
+                # number yet
+                invoice.avatax_compute_taxes(commit=False)
         res = super()._post()
```
atchuthan, line 305: res = super()._post() ==>  res = super(AccountMove, invoice)._post() such that we will call the super for each invoice in the for loop

atchuthan (changes requested):

account_avatax/models/account_move.py
dreispt: Actually, this super() call is not inside the for loop, so it is fine.

dreispt (approved):

account_avatax/models/account_move.py
dnplkndll: I did it that way to preserve the return value. If there is better style let me know. I just needed a quick fix for this week prod release!

account_avatax/models/account_move.py
```diff
@@ -287,24 +287,26 @@ def is_avatax_calculated(self):
...
-            self.avatax_compute_taxes(commit=False)
+        for invoice in self:
+            if invoice.is_avatax_calculated():
+                avatax_config = self.company_id.get_avatax_config_company()
+                if avatax_config and avatax_config.force_address_validation:
+                    for addr in [self.partner_id, self.partner_shipping_id]:
+                        if not addr.date_validation:
+                            # The Validate action will be interrupted
+                            # if the address is not validated
+                            return addr.button_avatax_validate_address()
```
atchuthan: @dreispt If the Avatax address validation is enabled and we use the bank statement to validate it, then the account.move will not be posted but the bank statement will be in posted state.

I think it is better to raise the Warning in this case instead of returning action

atchuthan (changes requested):

account_avatax/models/account_move.py
dreispt: OK for me.

atchuthan (approved):

bodedra:
@dnplkndll @dreispt @atchuthan IMO we should avoid tax re-computation after the post. Because there will be a penny off on the JE side. Do all tax computation and simply return super post() in the last.

dreispt:
@bodedra That sounds like a different scope from what was being fixed here.
You can go ahead and prepare a PR for that.

bodedra:
@dreispt You are right. I just reviewed entire method and see a gap.

It seems this patch haven't moved to v13.0. I will take care of it.

bodedra:
@dreispt PRs are ready for review.

- v13.0 https://github.com/OCA/account-fiscal-rule/pull/192 
- v14.0 https://github.com/OCA/account-fiscal-rule/pull/193

PR #192, opened 2021-05-03
[13.0][FIX] account_avatax: full refund, closes

bodedra:
This PR contains backport of https://github.com/OCA/account-fiscal-rule/pull/191 and address comment https://github.com/OCA/account-fiscal-rule/pull/191#issuecomment-831391292

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -1,6 +1,7 @@
 import logging
 
-from odoo import api, fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 from odoo.tests.common import Form
 
 _logger = logging.getLogger(__name__)
@@ -282,7 +283,7 @@ class AccountMove(models.Model):
                         if not addr.date_validation:
                             # The Validate action will be interrupted
                             # if the address is not validated
-                            return addr.button_avatax_validate_address()
+                            raise UserError(_("Avatax address is not validated!"))
                 # We should compute taxes before validating the invoice
                 # to ensure correct account moves
                 # However, we can't save the invoice because it wasn't assigned a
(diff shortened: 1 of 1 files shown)


dreispt:
@bodedra Can we have two separate PRs, one for the backport fix, and the otehr for the post() proposed fix?

bodedra:
@dreispt backport PR https://github.com/OCA/account-fiscal-rule/pull/194

bodedra:
@dreispt Both PRs affect same method. So I will propose post() fix once we merge backport. Sounds good?

dreispt:
OK

dreispt:
Regarding the post() fix, it is a bit trickier than that.
The naive solution is to have the first loop use `commit=True` on the Avatax call, and completely remove the second for loop.
This will do the same as the proposed solution, with half the API calls.

Now there is a risk if we do this: what if the Odoo `post()` fails?
The Avatax call was already made, so it is commitetd to Avatax.
But the Odoo invoice is rolled back, as is not stored in the system.
We now have an inconsistency.

The solution is for the second loop to use an API call changing the state of the Avatax draft document to committed, without recomputing taxes, as it is doing now.
I see no reason for the the second Avatax call to compute a different amount ("a penny off"), so there *should* not be a real problem here.
Still, this later implementation seems safer and also more efficient.

bodedra:
If Odoo `post()` fails, we should voided that transaction to Avatax.

bodedra:
I faced issue before, `Entity not found` during reset to draft invoice. It seems Odoo recorded transaction as a Avatax and Avatax doesn't found transaction.  I may be talking about different issue.

dreispt:
That's the symptom, but it usually is caused several Odoo test instances sharing the same Avatax service backend.

To set a document to committed use `avatax_config.commit_transaction(self.name, doc_type)`.
Example: https://github.com/OCA/account-fiscal-rule/blob/14.0/account_avatax/models/account_move.py#L226

PR #193, opened 2021-05-03
[14.0][FIX] post() super call after Avatax computation

bodedra:


bodedra:
Close this with ref https://github.com/OCA/account-fiscal-rule/pull/192#issuecomment-831454000 

I will reopen if needed.

PR #199, opened 2021-05-12
[14.0][FIX] account_avatax: Fix tax computation for multi-company and tax exemption on some detail taxes

mgosai:


account_avatax/models/account_tax.py
```diff
@@ -41,7 +42,22 @@ def get_avalara_tax(self, tax_rate, doc_type):
                 tax.name = self._get_avalara_tax_name(tax_rate, doc_type)
             return tax
         else:
-            tax = self.env.ref("account_avatax.avatax", raise_if_not_found=False)
+            # Get tax based on company
+            tax = self.search([
+                ("is_avatax", "=", True),
+                ("name", "=", "AVATAX"),
```
bodedra, line 48: instead of name search, can we use Avatax record XML id? (Ref https://github.com/OCA/account-fiscal-rule/blob/14.0/account_avatax/data/avalara_salestax_data.xml#L12)

bodedra:
Thanks @mgosai for fix this issue.

Instead of search by name, how about following approach:

- Add `Default Avatax` (default_avatax_config_id) many2one field at Company form.
- And simply check if `self.env.company.default_avatax_config_id`

account_avatax/models/account_tax.py
```diff
@@ -17,6 +17,7 @@ def _get_avalara_tax_domain(self, tax_rate, doc_type):
         return [
             ("amount", "=", tax_rate),
             ("is_avatax", "=", True),
+            ("company_id", "=", self.env.user.company_id and self.env.user.company_id.id),
```
dreispt: This is safe, AFAICT
```suggestion
            ("company_id", "=", self.env.user.company_id.id),
```

dreispt:
@bodedra I made changes. Can you test if this works as you expect?

dreispt:
@mgosai Can you test this on your end?

mgosai:
> @mgosai Can you test this on your end?

@dreispt Tested and works, also discovered another issue for product-exemption fetching wrong tax label. 
Exploring it more: Response from Avalara was giving accurate tax but the tax label that Odoo picks was wrong and hence results in confusion of sale order lines tax labels vs total order tax amount.
Issue:
image

After fix:
image

cc @bodedra

account_avatax/models/avatax_rest_api.py
```diff
@@ -305,7 +305,7 @@ def get_tax(
         # Enrich Avatax result with Odoo tax computation
         for line in result.get("lines", []):
             line["rate"] = (
-                round(sum(x["rate"] for x in line["details"]) * 100, 4)
+                round(sum(x["rate"] for x in line["details"] if "tax" in x and x["tax"] > 0) * 100, 4)
```
dreispt: Making it more Pythonic.
```suggestion
                round(sum(x["rate"] for x in line["details"] if x.get("tax")) * 100, 4)
```

I wish we had a test case for this line of code.
Isolating this bit of code in a helper function could make that easier to do.

dreispt:
@mgosai @bodedra 
We need to level up and start adding tests for fixes. Othewise we will see new fixes that break thing that were working before.
I did this myself for this tax exemption fix, here: https://github.com/OCA/account-fiscal-rule/pull/199/commits/97f17d7334d920aebf647ecf105e4041fc3db096

Please double check this changed version is still working as expected, and I'll merge.

dreispt (approved):

PR #200, opened 2021-05-19
[14.0][FIX] account_avatax,account_avatax_sale: avoid long field computations on module install

dreispt:
When installing on a database with many Account Moves or Sales Orders, it takes too much time and is interrupted by worker time limits.


bodedra (approved):
Code review. :+1:

bodedra:
@dreispt Should we take care of all compute fields on the Avalara with this approach or will this sufficient?

dreispt:
@bodedra I'm hoping this fixes the bottleneck. I'll ask the people that reported it to me for confirmation.

PR #204, opened 2021-05-28
[FIX] account_avatax: Get username & password from config

atchuthan:
if not found in the parameter set to False
https://github.com/OCA/account-fiscal-rule/pull/199 changed the way the AvaTaxRESTService __init__ is called.
cc: @dreispt @mgosai @bodedra @stephankeller @SodexisTeam


dreispt (approved):

dreispt:
Simple fix, fast tracking.

PR #206, opened 2021-06-14
[13.0][ADD] Account Fiscal Position Partner Type

ValentinVinagre:
Issue: https://github.com/OCA/l10n-spain/issues/1718
To Do:

- [X] Usage.rst
- [X] Description.rst
- [x] Tests


account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,115 @@
...
+from odoo import api, fields, models
+
+
+class AccountFiscalPosition(models.Model):
+    _inherit = "account.fiscal.position"
+
+    fiscal_position_type = fields.Selection(
+        selection=[("b2c", "Customer"), ("b2b", "Company")],
+        string="Type",
+        default=lambda self: self.env.company.default_fiscal_position_type,
```
RodrigoBM: is it possible to change this to a method?

default=lambda self: self._default_fiscal_position_type(),
    )

    @api.model
    def _default_fiscal_position_type(self):
        return self.env.company.default_fiscal_position_type

account_fiscal_position_partner_type/models/res_partner.py
```diff
@@ -0,0 +1,18 @@
...
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    fiscal_position_type = fields.Selection(
+        selection=[("b2c", "Customer"), ("b2b", "Company")],
+        string="Fiscal Position Type",
+        default=lambda self: self.env.company.default_fiscal_position_type,
```
RodrigoBM: is it possible to change this to a method?
```

        default=lambda self: self._default_fiscal_position_type(),
    )

    @api.model
    def _default_fiscal_position_type(self):
        return self.env.company.default_fiscal_position_type
```

account_fiscal_position_partner_type/models/account_fiscal_position.py
ValentinVinagre: Done

account_fiscal_position_partner_type/models/res_partner.py
ValentinVinagre: Done

colmeirin (approved):
Functional review in fresh instance.
Everything looks fine and working properly with #207

account_fiscal_position_partner_type/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+{
+    "name": "Account Fiscal Position Partner Type",
+    "version": "13.0.0.0.1",
+    "category": "Accounting & Finance",
+    "website": "https://github.com/OCA/account-fiscal-rule",
+    "author": "Sygel Technology," "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "development_status": "Production/Stable",
```
sergiocorato, line 13: This can be `Alpha` or `Beta`

pedrobaeza: If it meets the requirements (good test coverage), it can directly be this status.

RodrigoBM (approved):

account_fiscal_position_partner_type/__manifest__.py
sergiocorato: This module has no tests

pedrobaeza: Then it can't be indeed

ValentinVinagre: Hi @sergiocorato ,
The pr is in wip and there is a point that is to do tests that is not marked, I just uploaded the tests. If still the state has to be beta, I change it, I have no problem.

sergiocorato: Now is ok, thanks

sergiocorato (approved):
LGTM

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,119 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class AccountFiscalPosition(models.Model):
+    _inherit = "account.fiscal.position"
+
+    fiscal_position_type = fields.Selection(
+        selection=[("b2c", "Customer"), ("b2b", "Company")],
```
pedrobaeza:
```suggestion
        selection=[("b2c", "End customer (B2C)"), ("b2b", "Company (B2B)")],
```

account_fiscal_position_partner_type/models/res_company.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    default_fiscal_position_type = fields.Selection(
+        selection=[("b2c", "Customer"), ("b2b", "Company")],
+        string="Default Partner Fiscal Position Type",
+    )
```
pedrobaeza: For using the same definition as the original one:

```suggestion
    default_fiscal_position_type = fields.Selection(
        selection="_selection_fiscal_position_type`,
        string="Default Partner Fiscal Position Type",
    )
    
    def _selection_fiscal_position_type(self):
        field = "fiscal_position_type"
        return self.env["account.fiscal.position"].fields_get(allfields=[field])[field]["selection"]
```

account_fiscal_position_partner_type/models/res_partner.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    fiscal_position_type = fields.Selection(
+        selection=[("b2c", "Customer"), ("b2b", "Company")],
```
pedrobaeza: Same for the getting the selection.

RodrigoBM:
LGTM

account_fiscal_position_partner_type/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module allows to configure the customer type in partner and fiscal position forms.
```
pedrobaeza:
```suggestion
This module adds the concept of type applied to fiscal positions and partners. The idea is to categorize them for not applying fiscal positions that doesn't belong to the same type.
```

account_fiscal_position_partner_type/readme/USAGE.rst
```diff
@@ -0,0 +1,11 @@
+User can configure the default value according to the system company going to:
```
pedrobaeza: This first part should go to `CONFIGURE.rst` file.

account_fiscal_position_partner_type/views/account_fiscal_position.xml
```diff
@@ -0,0 +1,20 @@
...
+<odoo>
+    <record id="view_account_position_form_fiscal_position_type" model="ir.ui.view">
+        <field name="name">view.account.position.form.fiscal.position.type</field>
+        <field name="model">account.fiscal.position</field>
+        <field name="inherit_id" ref="account.view_account_position_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='auto_apply']" position="after">
+                <field
+                    name="fiscal_position_type"
+                    attrs="{'invisible': [('auto_apply', '!=', True)]}"
```
pedrobaeza, line 15: Why limiting to `auto_apply`? This can help in other use cases apart from OSS. That's why we put it as a separate module.

account_fiscal_position_partner_type/views/res_partner.xml
```diff
@@ -0,0 +1,20 @@
...
+<odoo>
+    <record id="view_partner_form_fiscal_position_type" model="ir.ui.view">
+        <field name="name">view.partner.form.fiscal.position.type</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='vat']" position="before">
+                <field
+                    name="fiscal_position_type"
+                    attrs="{'invisible': [('parent_id', '=', True)]}"
```
pedrobaeza, line 15: We should put this together the fiscal position, not the VAT.

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if not fpos:
+            # Fallback on catchall (no country, no group)
+            fpos = self.search(base_domain + null_country_dom, limit=1)
+        return fpos or False
+
+    @api.model
+    def get_fiscal_position(self, partner_id, delivery_id=None):
+        if not partner_id:
+            return self.env["account.fiscal.position"]
```
pedrobaeza: You shouldn't return early, but call super in that case. You can avoid the else branch and put the end the `super` call if the rest is not applying

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,119 @@
...
+                    delivery.state_id.id,
+                    delivery.zip,
+                    False,
+                    delivery.commercial_partner_id.fiscal_position_type,
+                )
+            if fp:
+                fp_id = fp.id
+        else:
+            fp_id = super(AccountFiscalPosition, self).get_fiscal_position(
+                partner_id, delivery_id
```
pedrobaeza: Don't change keyword arguments:

```suggestion
                partner_id, delivery_id=delivery_id
```

pedrobaeza (changes requested):
For getting the fiscal position, we are duplicating code for discovering the fiscal position, so there's the problem of diverging from core. Try this other approach:

* If the partner has a type, then we call super with a special context. For example, `fiscal_position_type=1`.
* We override the `_search` or `search` method of `account.fiscal.position`, and if the context is detected, the tuple `("fiscal_position_type", "=", context["fiscal_position_type"])` to the domain before calling `super`.

account_fiscal_position_partner_type/views/account_fiscal_position.xml
ValentinVinagre: Because the search logic that takes into account the 'fiscal_position_type' field does not apply if the 'auto_apply' field is not activated.

account_fiscal_position_partner_type/views/res_partner.xml
ValentinVinagre: I thought that, but if it is a field that will have to be configured every time a partner is created that is next to the 'VAT' field, I see it more correct, since it can be configured faster.

pedrobaeza: OK, you are right. The fiscal position will be auto-applied, so no configuration is done there.

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,119 @@
...
+            state_domain = [("state_ids", "=", state_id)]
+
+        domain_country = base_domain + [("country_id", "=", country_id)]
+        domain_group = base_domain + [("country_group_id.country_ids", "=", country_id)]
+
+        # Build domain to search records with exact matching criteria
+        fpos = self.search(domain_country + state_domain + zip_domain, limit=1)
+        # return records that fit the most the criteria, and fallback on
+        # less specific fiscal positions if any can be found
+        if not fpos and state_id:
```
RodrigoBM: I think this part is not correct.

- if not fpos and state_id: search with state_domain and null_zip_dom
- if not fpos and zipcode: search with null_state_dom and zip_domain
- if not fpos: search same domain_country but null_state_dom and  null_zip_dom

```
        if not fpos and state_id:
            fpos = self.search(domain_country + state_domain + null_zip_dom, limit=1)
        if not fpos and zipcode:
            fpos = self.search(domain_country + null_state_dom + zip_domain, limit=1)
        if not fpos:
            fpos = self.search(domain_country + null_state_dom + null_zip_dom, limit=1)

```

pedrobaeza: See https://github.com/OCA/account-fiscal-rule/pull/206#pullrequestreview-689503368 for the solution for not reimplementing the logic. DRY solution.

ValentinVinagre: I have already done the changes, I am checking that everything is correct and I do a push the changes. Good suggestion @pedrobaeza  had not fallen into doing so.

RodrigoBM: okey, that looks better

account_fiscal_position_partner_type/models/account_fiscal_position.py
pedrobaeza: This hasn't been changed yet

account_fiscal_position_partner_type/views/account_fiscal_position.xml
pedrobaeza: I think you shouldn't limit that. This is applying to the method `get_fiscal_position`, being called by auto-applied or whatever.

account_fiscal_position_partner_type/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,11 @@
+User can configure the default value according to the system company going to:
+   * Settings/Users & Companies/Companies.
```
pedrobaeza: Empty line needed and no indentation for proper RST syntax

account_fiscal_position_partner_type/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,11 @@
...
+   * Changing the value of the "Default Partner Fiscal Position Type" field.
+
+Create a new partner and define the "Fiscal Position Type" field.
```
pedrobaeza: I mentioned only the first part to be in `CONFIGURE.rst`. From here it should still go to `USAGE.rst`

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,67 @@
...
+            if delivery_id:
+                delivery = PartnerObj.browse(delivery_id)
+            else:
+                delivery = partner
+            fp_id = self.env["account.fiscal.position"]
+            # Only type has been configured
+            if (
+                delivery.commercial_partner_id.fiscal_position_type
+                and not delivery.property_account_position_id
+                and not delivery.commercial_partner_id.property_account_position_id
```
pedrobaeza: Isn't this data auto-propagated? I think this one is not needed.

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,67 @@
...
+    @api.model
+    def get_fiscal_position(self, partner_id, delivery_id=None):
+        if partner_id:
+            PartnerObj = self.env["res.partner"]
+            partner = PartnerObj.browse(partner_id)
+            # if no delivery use invoicing
+            if delivery_id:
+                delivery = PartnerObj.browse(delivery_id)
+            else:
+                delivery = partner
```
pedrobaeza:
```suggestion
            delivery = self.env["res.partner"].browse(delivery_id or partner_id)
```

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,67 @@
...
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        args = expression.AND(
+            (
+                args,
+                [
+                    (
+                        "fiscal_position_type",
+                        "=",
+                        self.env.context.get("fiscal_position_type", False),
```
pedrobaeza: Only apply this if `"fiscal_position_type" in self.env.context`

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,67 @@
...
+class AccountFiscalPosition(models.Model):
+    _inherit = "account.fiscal.position"
+
+    fiscal_position_type = fields.Selection(
+        selection=[("b2c", "End customer (B2C)"), ("b2b", "Company (B2B)")],
+        string="Type",
+        default=lambda self: self._default_fiscal_position_type(),
+    )
+
+    def _selection_fiscal_position_type(self):
```
pedrobaeza: This method is not needed here

pedrobaeza (changes requested):

account_fiscal_position_partner_type/models/account_fiscal_position.py
ValentinVinagre: hummm if I apply it only when the 'fiscal_position_type' is in self.env.context. When the partner doesn't have a 'tax_position_type' configured, the search takes into account the tax position configured with 'tax_position_type' and it shouldn't, that's why I do it this way.

pedrobaeza: Then always pass the context when coming from the other method, but what we don't want is that this limited search restricts other regular searches. If not, you won't be able to see all the fiscal position types. Do you understand me?

ValentinVinagre: Now yes :). Sorry, hard day.

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,54 @@
...
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        if self.env.context.get("fiscal_position_type"):
+            args = expression.AND(
+                (
+                    args,
+                    [
+                        (
+                            "fiscal_position_type",
+                            "=",
+                            self.env.context.get("fiscal_position_type", False),
```
pedrobaeza, line 31: For shorter code:
```suggestion
                            self.env.context["fiscal_position_type"],
```

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,54 @@
...
+        default=lambda self: self._default_fiscal_position_type(),
+    )
+
+    @api.model
+    def _default_fiscal_position_type(self):
+        return self.env.company.default_fiscal_position_type
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        if self.env.context.get("fiscal_position_type"):
```
pedrobaeza, line 23:
```suggestion
        if "fiscal_position_type" in self.env.context:
```

pedrobaeza (changes requested):

account_fiscal_position_partner_type/models/account_fiscal_position.py
pedrobaeza: Sorry, I get what you want to achieve this way. It's OK then.

pedrobaeza (approved):

account_fiscal_position_partner_type/views/account_fiscal_position.xml
ValentinVinagre: I've done it this way because odoo in the '_get_fpos_by_region' function always searches with the 'auto_apply' field set to true.

https://github.com/odoo/odoo/blob/17e19c6bf787fd2bf77476ffcfa1d934058aa7ec/addons/account/models/partner.py#L126-L130

If the field is displayed without being used in the search, it could lead to think that the module does not work.

HaraldPanten (approved):
Ok,

Let's finish with this first approach and later we could improve it 👍

account_fiscal_position_partner_type/views/account_fiscal_position.xml
pedrobaeza: OK, seems correct to me. On the way to merging!

ValentinVinagre:
@pedrobaeza thanks a lot for the review :)

RodrigoBM:
add cherry pick https://github.com/OCA/account-fiscal-rule/pull/210

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+
+    @api.model
+    def get_fiscal_position(self, partner_id, delivery_id=None):
+        fiscal_type = False
+        if partner_id:
+            delivery = self.env["res.partner"].browse(delivery_id or partner_id)
+            # Only type has been configured
+            if (
+                delivery.fiscal_position_type
```
dreispt, line 47: @ValentinVinagre Does it make sense to fallback to the Company Default, when not set on the Partner?

PR #207, opened 2021-06-14
[13.0][ADD]L10n EU OSS

ValentinVinagre:
Issue: https://github.com/OCA/l10n-spain/issues/1718
Depend: https://github.com/OCA/account-fiscal-rule/pull/206
To Do:

- [X] Usage.rst
- [X] Description.rst
- [X] Tests

PR #209, opened 2021-06-17
11.0 mig l10n eu oss account fiscal position partner type 

RodrigoBM:
add cherry pick 
https://github.com/OCA/account-fiscal-rule/pull/206
https://github.com/OCA/account-fiscal-rule/pull/207

Issue: https://github.com/OCA/l10n-spain/issues/1718


ValentinVinagre:
Hi @RodrigoBM ,
It would be better to separate these pr (one by module).

PR #212, opened 2021-06-22
[13.0][ADD] account_fiscal_position_partner_domain

AdriaGForgeFlow:
This PR adds the modules _account_fiscal_position_get_fpos_by_region_hook_ and _account_fiscal_position_partner_domain_ to v13.

The module _account_fiscal_position_get_fpos_by_region_hook_ introduces a hook in the function _get_fpos_by_region to be used by the second module

The module _account_fiscal_position_partner_domain_ allows to define a partner domain in the Fiscal Position that will be used in the automatic
detection evaluation as follows:

* First, the system will look for a Fiscal Position using the Odoo standard approach
* Second, the system will try to look for a Fiscal Position applying the partner domain
* The system will return one of the found Fiscal Positions (if any) giving priority to the one that has a partner domain set

cc ~ @ForgeFlow


account_fiscal_position_partner_domain/models/account_fiscal_position.py
```diff
@@ -0,0 +1,116 @@
...
+        if not fpos:
+            fpos = self.search(domain_group + null_state_dom + null_zip_dom)
+
+        if not fpos:
+            # Fallback on catchall (no country, no group)
+            fpos = self.search(base_domain + null_country_dom)
+
+        # Evaluate domain in the Fiscal Positions that have been found
+        for fiscal_position in fpos:
+            partner_domain = literal_eval(fiscal_position.partner_domain)
```
JordiMForgeFlow, line 72: I think that here it would need to check if there is actually a partner_domain set in the fiscal position, as if it is empty ([]) it matches all records.

AdriaGForgeFlow: I'm doing that here https://github.com/OCA/account-fiscal-rule/pull/212/files#diff-6cb6576d850e90ddac4f46fd15fcd258a9634a2cd4e06b8f21058f865ada2344R34

The base_domain contains `[("partner_domain", "!=", False)]`, so all the Fiscal Positions found should have a partner_domain set.

JordiMForgeFlow: I see, the problem is that if you add some filters in the domain and then you remove them, the partner_domain is left with [], so that is when you have the issue.

JordiMForgeFlow: image
image

AdriaGForgeFlow: Nice catch :+1: 

I have updated the base_domain so it also filters out those Fiscal Position with partner_domain = [ ]

JordiMForgeFlow (approved):

account_fiscal_position_partner_domain/models/account_fiscal_position.py
```diff
@@ -0,0 +1,116 @@
...
+    _inherit = "account.fiscal.position"
+
+    partner_domain = fields.Char()
+
+    def _get_base_domain(self, vat_required, company_id, partner_domain=False):
+        base_domain = super()._get_base_domain(vat_required, company_id)
+        extra_domain = (
+            [("partner_domain", "!=", False), ("partner_domain", "!=", "[]")]
+            if partner_domain
+            else [("partner_domain", "=", False), ("partner_domain", "=", "[]")]
```
JordiMForgeFlow:
```suggestion
            else ['|', ("partner_domain", "=", False), ("partner_domain", "=", "[]")]
```

pedrobaeza:
This is very similar to what we have done at https://github.com/OCA/account-fiscal-rule/tree/13.0/account_fiscal_position_partner_type

pedrobaeza:
Check also the technique used there for not doing such "dirty hook".

PR #214, opened 2021-06-24
[14.0][MIG] Account Fiscal Position Partner Type: Migrated into v14

Vimal-OSI:


account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -11,24 +11,20 @@ class AccountFiscalPosition(models.Model):
...
     )
 
-    @api.model
-    def _default_fiscal_position_type(self):
-        return self.env.company.default_fiscal_position_type
-
     @api.model
     def search(self, args, offset=0, limit=None, order=None, count=False):
-        if self.env.context.get("fiscal_position_type"):
+        if "fiscal_position_type" in self.env.context:
```
dreispt: This changes behavior if contact has {"fiscal_position_type": False}
Don't change it please.

dreispt (changes requested):

account_fiscal_position_partner_type/models/account_fiscal_position.py
Vimal-OSI: Done the suggested change.
Thank you.

dreispt (approved):

PR #217, opened 2021-06-29
[13.0][IMP] l10n_eu_oss: tax groups are created and searched for the new taxes.

ValentinVinagre:
tax groups are created and searched for the new taxes.

### DIFF
--- a/l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
+++ b/l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
@@ -90,6 +90,9 @@ class L10nEuOssWizard(models.TransientModel):
         comodel_name="account.tax", string="Second Super Reduced Tax"
     )
 
+    def _prepare_tax_group_vals(self, rate):
+        return {"name": _("OSS VAT %s%%") % rate}
+
     def _prepare_repartition_line_vals(self, original_rep_lines):
         return [
             (
@@ -108,7 +111,7 @@ class L10nEuOssWizard(models.TransientModel):
             for line in original_rep_lines
         ]
 
-    def _prepare_tax_vals(self, country_id, tax_id, rate):
+    def _prepare_tax_vals(self, country_id, tax_id, rate, tax_group):
         return {
             "name": _("OSS for EU to %(country_name)s: %(rate)s")
             % {"country_name": country_id.name, "rate": rate},
@@ -124,6 +127,7 @@ class L10nEuOssWizard(models.TransientModel):
             "oss_country_id": country_id.id,
             "company_id": self.company_id.id,
             "price_include": self.price_include_tax,
+            "tax_group_id": tax_group.id,
             "sequence": 1000,
         }
 
@@ -161,6 +165,7 @@ class L10nEuOssWizard(models.TransientModel):
     def generate_eu_oss_taxes(self):
         oss_rate = self.env["oss.tax.rate"]
         account_tax = self.env["account.tax"]
+        account_tax_group = self.env["account.tax.group"]
         selected_taxes = []
         fpos_obj = self.env["account.fiscal.position"]
         # Get the taxes configured in the wizard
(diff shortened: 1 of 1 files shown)


l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
```diff
@@ -90,6 +90,9 @@ def _default_todo_country_ids(self):
         comodel_name="account.tax", string="Second Super Reduced Tax"
     )
 
+    def _prepare_tax_group_vals(self, rate):
+        return {"name": _("OSS VAT {}{}".format(rate, "%"))}
```
pedrobaeza: Use old replacing system (`%s`) and don't include in the translatable part the replacing itself:
```suggestion
        return {"name": _("OSS VAT %s%%") % rate}
```

l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
```diff
@@ -194,8 +199,15 @@ def generate_eu_oss_taxes(self):
                         limit=1,
                     )
                     if not tax_dest_id:
+                        tax_group_id = account_tax_group.search(
+                            [("name", "=", _("OSS VAT {}{}".format(rate, "%")))]
```
pedrobaeza: Don't recreate the same group name, but call `self._prepare_tax_group_vals()["name"]`

l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
```diff
@@ -194,8 +199,15 @@ def generate_eu_oss_taxes(self):
...
+                        )
+                        if not tax_group_id:
+                            tax_group_id = account_tax_group.create(
+                                self._prepare_tax_group_vals(rate)
+                            )
                         tax_dest_id = account_tax.create(
-                            self._prepare_tax_vals(country, tax, rate)
+                            self._prepare_tax_vals(country, tax, rate, tax_group_id)
```
pedrobaeza: Don't change the arguments, but do the search/create inside the `_prepare_tax_vals` method IMO

pedrobaeza (changes requested):

l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
ValentinVinagre: I don't want to search / create the tax group in _prepare_tax_vals, I think this method should only add the fields to the dictionary, how do you see it?

pedrobaeza: OK, I don't have right now an strong opinion on this. It's for being more self-contained.

l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
```diff
@@ -108,7 +111,7 @@ def _prepare_repartition_line_vals(self, original_rep_lines):
             for line in original_rep_lines
         ]
 
-    def _prepare_tax_vals(self, country_id, tax_id, rate):
+    def _prepare_tax_vals(self, country_id, tax_id, rate, tax_group_id):
```
pedrobaeza: Please don't use the suffix `_id` when you are passing a recordset.

```suggestion
    def _prepare_tax_vals(self, country_id, tax_id, rate, tax_group):
```

l10n_eu_oss/wizard/l10n_eu_oss_wizard.py
```diff
@@ -194,8 +199,15 @@ def generate_eu_oss_taxes(self):
                         limit=1,
                     )
                     if not tax_dest_id:
+                        tax_group_id = account_tax_group.search(
```
pedrobaeza: The same here about suffix.

pedrobaeza (approved):

ValentinVinagre:
thanks for review @pedrobaeza

HaraldPanten (approved):
Functional review 👍

Jortolsa-S73:
Today we will publish the PR with the migration of this module to 14.0.

ValentinVinagre:
> Today we will publish the PR with the migration of this module to 14.0.

Perfect @Jortolsa . Can you say it in https://github.com/OCA/account-fiscal-rule/issues/138 ? Thanks

PR #224, opened 2021-07-06
[8.0][MIG] account_fiscal_position_partner_type

RodrigoBM:
add cherry pick #206

Issue: OCA/l10n-spain#1718


account_fiscal_position_partner_type/models/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- encoding: utf-8 -*-
+# Copyright 2021 Valentin Vinagre <valentin.vinagre@sygel.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import res_company
+from . import res_partner
+from . import account_fiscal_position
```
pedrobaeza, line 6: I think this one should be before the others, as it's used for getting the selection. Anyways, it can be possible that the used trick doesn't serve for this version.

RodrigoBM: I think its Ok. I needed to add a hook to pass an error when installing the module.

account_fiscal_position_partner_type/hooks.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 FactorLibre - Rodrigo Bonilla <rodrigo.bonilla@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def pre_init_hook(cr):
+    cr.execute(
+        """
+ALTER TABLE res_company
+ADD COLUMN default_fiscal_position_type character varying DEFAULT ''
```
pedrobaeza: This is incorrect, as it's putting a default for ALL new values, and the default is not correct (it should be null, not empty string). Anyways, this is not needed at all. See below.

account_fiscal_position_partner_type/models/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- encoding: utf-8 -*-
+# Copyright 2021 Valentin Vinagre <valentin.vinagre@sygel.es>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from . import account_fiscal_position
+from . import res_partner
+from . import res_company
```
pedrobaeza: Putting this order works perfectly:

```suggestion
from . import res_company  # This one should be first
from . import account_fiscal_position
from . import res_partner
```

pedrobaeza (changes requested):

pedrobaeza:
@RodrigoBM can you make the changes and we merge?

account_fiscal_position_partner_type/models/__init__.py
RodrigoBM: @pedrobaeza The problem is that Travis doesn't pass with this configuration. I need to create a hook so that Travis doesn't generate an error, although I haven't managed to understand the final reason for the error.

pedrobaeza: Let me check. It's a collision with other module, as installed alongside it's working.

pedrobaeza:
I have found that demo data loaded via YAML leads to inconsistencies, so I have deactivated the conflicting ones here: b7c14a13a1dd1512c96cd71e2ef9edb13f8d3db2

As I can't push to this branch, as it's an organization one, I'm superseding this in #228

PR #227, opened 2021-07-09
[13.0] [IMP] Improved string in fiscal position type

ValentinVinagre:
IMP: added 'self-employed' concept in fiscal_partner_type string to consider these use cases

### DIFF
--- a/account_fiscal_position_partner_type/models/account_fiscal_position.py
+++ b/account_fiscal_position_partner_type/models/account_fiscal_position.py
@@ -9,7 +9,10 @@ class AccountFiscalPosition(models.Model):
     _inherit = "account.fiscal.position"
 
     fiscal_position_type = fields.Selection(
-        selection=[("b2c", "End customer (B2C)"), ("b2b", "Company (B2B)")],
+        selection=[
+            ("b2c", "End customer (B2C)"),
+            ("b2b", "Company/Self-employed (B2B)"),
+        ],
         string="Type",
         default=lambda self: self._default_fiscal_position_type(),
     )


ValentinVinagre:
@HaraldPanten

account_fiscal_position_partner_type/models/account_fiscal_position.py
```diff
@@ -20,7 +20,7 @@ def _default_fiscal_position_type(self):
 
     @api.model
     def search(self, args, offset=0, limit=None, order=None, count=False):
-        if self.env.context.get("fiscal_position_type"):
+        if "fiscal_position_type" in self.env.context:
```
pedrobaeza: This is exactly what I suggested in https://github.com/OCA/account-fiscal-rule/pull/206/files#r656460329, but I analyzed it, and determined that it's OK as it is for only acting if a explicit type is given.

HaraldPanten: We know, Pedro, but we found some functional use cases where it should be specified when empty.

For example, for customers that were already created in the database it was auto-applying (calculating) fiscal position OSS rules in SO & Invoices but this one shouldn't be applied.

pedrobaeza: Please specify whole detailed steps to reproduce the problem to judge.

ValentinVinagre: Updated on main comment.

pedrobaeza:
I don't consider such case as valid: you have to explicitly set a partner type. No match should be done over 2 empty values (on fiscal position and on partner), as this can lead to unexpected matches due to a combination of not defining properly things. For me this should be closed, and maybe clarify in the README that a value must be explicitly provided on partners.

What you try to do is to set a default fiscal position, when that thing doesn't exist: or you have B2C, or B2B, not empty case, and as such, you have to fill existing fiscal positions with `B2B` for the rest of the cases, and partners not being B2C the same.

ValentinVinagre:
> I don't consider such case as valid: you have to explicitly set a partner type. No match should be done over 2 empty values (on fiscal position and on partner), as this can lead to unexpected matches due to a combination of not defining properly things. For me this should be closed, and maybe clarify in the README that a value must be explicitly provided on partners.
> 
> What you try to do is to set a default fiscal position, when that thing doesn't exist: or you have B2C, or B2B, not empty case, and as such, you have to fill existing fiscal positions with `B2B` for the rest of the cases, and partners not being B2C the same.

I don't agree that this should be closed.
The field is not mandatory, so it should allow you to search for any of the options indicated in the client and in the tax positions.
If, when leaving the customer field empty, a search is made on all tax positions without taking into account the type, this can lead to errors.

pedrobaeza:
An empty partner type expresses that no matter the fiscal position, not that the fiscal position should be explicitly False as well.

ValentinVinagre:
> An empty partner type expresses that no matter the fiscal position, not that the fiscal position should be explicitly False as well.

hummm But if the field is empty the system should not take into account the tax positions typified especially for the types 'b2c' / 'b2b' / ...

pedrobaeza:
As said, for me it means no preference, so the priority set on them will decide.

ValentinVinagre:
> As said, for me it means no preference, so the priority set on them will decide.

I understand that we have different views on this. We will do it privately for clients then. I have deleted the FIX and left the IMP.

pedrobaeza:
Others can join the debate and the majority should decide. Give your opinion @dreispt @RodrigoBM @angelmoya @OSevangelist

HaraldPanten:
We''ll just keep the IMP, as OSS rules shouldn't be applied for self-employed and it clarifies the new registered users selection options.

Maybe we could purpose the other FIX if others think it's useful. 👍

HaraldPanten (approved):

pedrobaeza (approved):

dreispt:
I think this change goes into an unneeded technicality, and probably is inaccurate in other jurisdictions:
the relevant criteria is something if the entity is a VAT taxable.
So I don't think this change is an improvement, the "B2B" there is enough the express these cases.

pedrobaeza:
@dreispt the debate was about the now removed other part.

dreispt:
PS: I don't think this was a priority fix worthy of a fast-track merge, and should have been given the standard review window.

pedrobaeza:
> PS: I don't think this was a priority fix worthy of a fast-track merge, and should have been given the standard review window.

Yes, I agree in this case @HaraldPanten has gone too quick

HaraldPanten:
@dreispt @pedrobaeza 

We suggested this change as we were having some issues with some clients. Some of them categorized themselves as customers (but they are self employed) instead of companies as they were misunderstanding the concepts.

Some of them were complaining because of theses VAT changes. They asked how come was it possible to have these taxes applied as they were self employed with a valid VIES. (One of them in Portugal, by the way)

As OSS is a global EU legislation, we thought that this IMP should have been done for everyone and not only for us.

We'll keep your suggestions in mind for next changes.

dreispt:
> Some of them were complaining because of theses VAT changes. They asked how come was it possible to have these taxes applied as they were self employed with a valid VIES. (One of them in Portugal, by the way)

For the specific case of Portugal, I'm aware that self employed are allowed to not charge VAT if their year revenue is under a threshold (10k€? I would need to check to be sure). Above that, you need to collect VAT as any company does.

So, the self-employed in general qualifies as "business", but there are VAT exemptions that small self-employed operation can benefit from.

Additionally, since the VAT is now collected for the country where the consumer is, I wonder if this Portuguese Tax Authority exemption applies at all, and there is a chance that the VAT would need to be collected anyway. This is pretty new and a tax specialist would need to be consulted.

PR #229, opened 2021-07-19
[FIX] account_product_fiscal_classification : force_company when writing fiscal information

legalsylvain:
in multicompany context, writing with sudo, without with_context(force_company )change the current company.
So computed fields based on company dependent fields will fail
(for exemple computation of theoritical_price, in the module product_margin_classification in OCA/margin-analysis)

CC : @quentinDupont  apps/deck/#/board/144/card/1722


quentinDupont:
use in prod since 3 months :heavy_check_mark:

PR #231, opened 2021-08-03
[FIX] account_avatax: tax copy method overrides the default dict

bodedra:
PR Description:

- In v14, (https://github.com/odoo/odoo/blob/14.0/addons/account/models/account_tax.py#L185) copy method in account/models/account_tax.py overrides the default dict passed and resets name. So, setting the name explicitly in the connector module.

### DIFF
--- a/account_avatax/models/account_tax.py
+++ b/account_avatax/models/account_tax.py
@@ -48,6 +48,8 @@ class AccountTax(models.Model):
                 "name": self._get_avalara_tax_name(tax_rate, doc_type),
             }
             tax = tax_template.sudo().copy(default=vals)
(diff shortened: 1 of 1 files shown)


bodedra:
cc @b-kannan @dreispt @atchuthan 

Thank you @b-kannan for fix this issue.

atchuthan (approved):

dreispt:
Thank you for the details @bodedra 
I feel there is a bug in the Odoo code - it should only set name if it is not already provided in the default dict.
It would be nice if an upstream PR could be proposed.

dreispt (approved):

PR #232, opened 2021-08-04
[14.0] account_avatax_exemption_base: change license to LGPL

atchuthan:
cc: @SodexisTeam


dnplkndll (approved):
functionally works well on US exemption management

odoo-sh:
Did QA and it is working well.

account_avatax_exemption_base/__manifest__.py
```diff
@@ -7,16 +7,15 @@
     """,
     "website": "https://github.com/OCA/account-fiscal-rule",
     "author": "Sodexis, Odoo Community Association (OCA)",
-    "license": "AGPL-3",
+    "license": "LGPL-3",
     "depends": [
-        "account_avatax",
-        "account_avatax_sale",
+        "mail",
+        "account",
```
dreispt, line 13: I'm quite sure this must depend on "account_avatax".
In that case, "mail" and "account" dependencies will not be explicitly needed.

dreispt (changes requested):
I have a concern with module dependencies. Can you check?

account_avatax_exemption_base/__manifest__.py
atchuthan: This module doesn't have direct relation to account_avatax.  Also, we are trying to make this module as LGPL license such that we can add further customization of enterprise module such as  sign and documents related to account_avatax_exemption in the future.

atchuthan: @dreispt please let us know if the above answer resolves your issue

amkarthik:
Looks good to me

amkarthik (approved):

dreispt (approved):

PR #233, opened 2021-08-04
[14.0] account_avatax: Add log to identify no. of call each day via email

atchuthan:
cc: @SodexisTeam


account_avatax/security/ir.model.access.csv
```diff
@@ -7,3 +7,4 @@ access_exemption_code_manager,exemption.code.manager,model_exemption_code,accoun
 access_exemption_code_employee,exemption.code.employee,model_exemption_code,base.group_user,1,0,0,0
 access_avalara_salestax_ping,avalara_salestax_ping,model_avalara_salestax_ping,base.group_user,1,1,1,1
 access_avalara_salestax_address_validate,avalara_salestax_address_validate,model_avalara_salestax_address_validate,base.group_user,1,1,1,1
+access_avatax_log,avatax_log,model_avatax_log,base.group_user,1,1,1,1
```
dreispt, line 10: The is written using sudo() do this is not needed for regular users.
I would change this to read only access to the Invoicing User, and maybe full access (include delete) to Settings users.

account_avatax/models/avatax_rest_api.py
```diff
@@ -319,6 +319,16 @@ def get_tax(
 
         response = self.client.create_or_adjust_transaction(data)
         result = self.get_result(response, ignore_error=ignore_error)
+        self.config.env["avatax.log"].sudo().create(
```
dreispt, line 322: This is a good feature - we needed a log on the Odoo side.
But I would only do this if the logging is enabled or there are notification users configured.

account_avatax/models/res_company.py
```diff
@@ -26,3 +27,37 @@ def get_avatax_config_company(self):
...
+            if calls
+            else False,
+        )
+        return res
+
+    def set_values(self):
+        res = super(Settings, self).set_values()
+        ICPSudo = self.env["ir.config_parameter"].sudo()
+        set_param = "account_avatax.avatax_api_call_notification_ids"
+        ICPSudo.set_param(set_param, self.avatax_api_call_notification_ids.ids)
```
dreispt, line 62: As a courtesy, should this also enable/disable the scheduled action?

dreispt (approved):

dreispt:
@atchuthan This is in draft state, are you planning to move it to a full PR?

dnplkndll:
@dreispt  this is a workaround for the lack of a counter api side. It would be great to pull from the avatax side if you see that feature become available. This will catch odoo automation issues causing excess api calls, not going to catch the same on external ecommerce apps accessing the same account for exemption access.

PR #235, opened 2021-08-10
[REF] adopting account_product_fiscal_classification and account_product_fiscal_classification_test

legalsylvain:
trivial PR.

Adopting two modules.

CC : recent contributors : @rvalyi, @PierrickBrun

### DIFF
--- a/account_product_fiscal_classification/__manifest__.py
+++ b/account_product_fiscal_classification/__manifest__.py
@@ -9,6 +9,7 @@
     "version": "14.0.1.0.0",
     "category": "Accounting",
     "author": "Akretion,GRAP,La Louve,Odoo Community Association (OCA)",
+    "maintainers": ["legalsylvain"],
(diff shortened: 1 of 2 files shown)


rvalyi (approved):
Thank you @legalsylvain. As for now in Brazil we abandonned this way for now and assumed to develop a specific model which ends up being much easier to set up and maintain for us (because it matches the fiscal data modelling more closely, the data is then more normalized for us in Brazil if you like). At least for the near future. BTW I have was writting the explanations why in the l10n_br_fiscal module README, one can read of curious, PR in process: https://github.com/akretion/l10n-brazil/tree/12.0-l10n_br_fiscal-readme2/l10n_br_fiscal#conceito-de-documento-fiscal

PierrickBrun (approved):

legalsylvain:
@rvalyi, could you merge it ?

Kind regards.

PR #236, opened 2021-08-12
[8.0][FIX] l10n_eu_oss

RodrigoBM:
In v8 it is necessary to create account_tax_code to assign them to taxes. If account_tax_code are not assigned to the taxes, they will not be passed to the accounting entries.

### DIFF
--- a/l10n_eu_oss/__openerp__.py
+++ b/l10n_eu_oss/__openerp__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "L10n EU OSS",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.1.0",
     "category": "Accounting & Finance",
     "website": "https://github.com/OCA/l10n-spain",
     "author": "Sygel Technology," "Odoo Community Association (OCA)",
--- a/l10n_eu_oss/migrations/8.0.1.1.0/post-migration.py
+++ b/l10n_eu_oss/migrations/8.0.1.1.0/post-migration.py
@@ -1,0 +1,52 @@
+# -*- encoding: utf-8 -*-
+# Copyright 2021 Rodrigo Bonilla <rodrigo.bonilla@factorlibre.com>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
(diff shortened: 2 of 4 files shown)


RodrigoBM:
@pedrobaeza

l10n_eu_oss/hooks.py
```diff
@@ -0,0 +1,57 @@
...
+
+from openerp import api, SUPERUSER_ID, _
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(cr, registry):
+    with api.Environment.manage():
+        env = api.Environment(cr, SUPERUSER_ID, {})
```
pedrobaeza: This shouldn't be a hook, but a migration script.

l10n_eu_oss/hooks.py
```diff
@@ -0,0 +1,57 @@
...
+                account_tax_id.write({
+                    "tax_code_id": code_c.id,
+                    "tax_sign": 1,
+                    "ref_tax_code_id": code_c.id,
+                    "ref_tax_sign": -1,
+                })
+
+
+def _upgrade_tax_code(env, account_tax_id, type):
+    account_tax_code = env["account.tax.code"]
```
pedrobaeza: Reuse existing code instead of duplicating it.

pedrobaeza (changes requested):

l10n_eu_oss/hooks.py
RodrigoBM: Done

l10n_eu_oss/hooks.py
RodrigoBM: Done, it's my first migrations, I hope it is correct. :)

RodrigoBM:
Se cambia el nombre del commit

pedrobaeza:
As I can't push to your branch, being an organization one, I have pushed directly to the branch with your commit 529e43650279098e1bde163618eaeca0d70960cc + an extra one for finishing the work: c1cdb1c2f05c39b39c86ede7d8aaa0fa5cf04322.

Thus, I close this.

PR #237, opened 2021-09-01
[14.0] account_avatax: Add Taxoverride for Credit notes(Refund)

atchuthan:
cc: @SodexisTeam


account_avatax/models/account_move.py
```diff
@@ -187,6 +192,9 @@ def _avatax_compute_tax(self, commit=False):
         if not avatax_config:
             # Skip Avatax computation if no configuration is found
             return
+        avatax_line_override = False
+        if self.avatax_amt_line_override and self.move_type == "out_refund":
+            avatax_line_override = True
```
dreispt: Simpler code:
```suggestion
        avatax_line_override = self.avatax_amt_line_override and self.move_type == "out_refund"
```

account_avatax/models/account_move.py
```diff
@@ -226,6 +235,9 @@ def _avatax_compute_tax(self, commit=False):
             avatax_config.commit_transaction(self.name, doc_type)
             return tax_result
 
+        if self.avatax_amt_line_override and self.move_type == "out_refund":
+            return
+
         if self.state == "draft":
```
dreispt: Simpler code: 
```suggestion
        if self.state == "draft" and not avatax_line_override:
```

account_avatax/models/account_move.py
```diff
@@ -407,11 +421,21 @@ def create(self, vals):
             record.avatax_compute_taxes()
         return record
 
+    def action_reverse(self):
+        action = super().action_reverse()
+        avatax_tax_type = self.invoice_line_ids.filtered(lambda t: t.avatax_tax_type)
+        action["context"] = {
+            "default_avatax_amt_line_override": self.avatax_amt_line_override,
+            "hide_override": 1 if avatax_tax_type else 0,
+        }
```
dreispt: I'm not sure it is safe to override the action context :-/
```suggestion
        action["context"] = action.get("context", {}).update(
            {
                "default_avatax_amt_line_override": self.avatax_amt_line_override,
                "hide_override": 1 if avatax_tax_type else 0,
            }
        )
```

account_avatax/models/account_move.py
```diff
@@ -465,6 +489,9 @@ def _avatax_prepare_line(self, sign=1, doc_type=None):
         amount = sign * line._get_avatax_amount()
         if line.quantity < 0:
             amount = -amount
+        avatax_amt = 0.0
+        if line.move_id.move_type == "out_refund":
```
dreispt, line 502: Also consider `avatax_amt_line_override`?
```suggestion
        if line.avatax_amt_line_override and line.move_id.move_type == "out_refund":
```

account_avatax/views/account_move_view.xml
```diff
@@ -14,6 +14,9 @@
                     attrs="{'invisible': [('state', 'not in', ['draft'])]}"
                 />
             </button>
+            <xpath expr="//button[@name='action_reverse']" position="attributes">
+                <attribute name="context">{'avatax_amt_line_override': True}</attribute>
+            </xpath>
```
dreispt: Why is this? We should be able to reverse invoices without using Avatax override.

account_avatax/views/account_move_view.xml
```diff
@@ -32,6 +35,10 @@
                     name="invoice_doc_no"
                     attrs="{'invisible': [('move_type','!=','out_refund')]}"
                 />
+                <field
+                    name="avatax_amt_line_override"
+                    groups="account_avatax.access_avatax_override_group"
```
dreispt, line 37: Shouldn't this field be visible only for refunds?
Doing so, the Python code could be simplified to rely only on this flag, without checking `move_type == "out_refund"`

In fact, I had requests to allow for some lines to use Odoo taxes, for cases where there were special taxes not handled by Avatax. Doing this would be a step towards supporting that.

account_avatax/models/account_move.py
```diff
@@ -226,6 +235,9 @@ def _avatax_compute_tax(self, commit=False):
             avatax_config.commit_transaction(self.name, doc_type)
             return tax_result
 
+        if self.avatax_amt_line_override and self.move_type == "out_refund":
```
dreispt: Can you confirm that the Invoice total tax amount is correct when you have some lines with Odoo TAx and other lines with Avalara tax?

dreispt (approved):

PR #238, opened 2021-09-23
[13.0][ADD] account_fiscal_position_autodetect_optional_vies: This module adds vat VIES validation to fiscal positions.

victoralmau:
This module adds vat VIES validation to fiscal positions.

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30631


account_fiscal_position_autodetect_optional_vies/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
+Create a new partner and define Spain in country field.
+
+Create a new Fiscal Position, with the following configuration:
+
+* 'Detect Automatically' -> checked.
+* 'Vat required' -> checked.
+* 'Vat VIES required' -> checked.
+* 'Country group' -> Europe.
```
dreispt: Question: as far as I can see, this is already the default behavior - VIES check only applies to EU countries.
What is the benefit of this feature?

victoralmau: Sorry for the confusion. It is not necessary to define any group of countries or country in the fiscal position, it will only be applied if the contact has a valid VIES.

dreispt: So 'Vat VIES required' means 'Vat was VIES validated'?  - Fiscal Position only applies if VAT passed VIES validation.

victoralmau: Exactly, I think it is better to change the string so that it is better understood. Now I will change it.

joao-p-marques (approved):

account_fiscal_position_autodetect_optional_vies/models/account_fiscal_position.py
```diff
@@ -0,0 +1,43 @@
...
+    _inherit = "account.fiscal.position"
+
+    vat_vies_required = fields.Boolean(
+        string="Vat was VIES validated",
+        help="Apply only if VAT passed VIES validation.",
+    )
+
+    @api.model
+    def search(self, args, offset=0, limit=None, order=None, count=False):
+        if self.env.context.get("vat_vies_required"):
```
pedrobaeza:
```suggestion
        if "vat_vies_required" in self.env.context:
```

account_fiscal_position_autodetect_optional_vies/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+This module adds vat VIES validation to fiscal positions.
```
pedrobaeza:
```suggestion
This module adds vat VIES validation to fiscal positions auto detection system.
```

account_fiscal_position_autodetect_optional_vies/readme/USAGE.rst
```diff
@@ -0,0 +1,9 @@
+Create a new partner and define some country.
```
pedrobaeza: The first two steps should go in CONFIGURE.rst, and please put them with the bullet list

pedrobaeza (changes requested):

account_fiscal_position_autodetect_optional_vies/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,7 @@
+- Create a new partner and define some country.
```
pedrobaeza, line 1: This is usually put as numbered list:
```suggestion
#. Create a new partner and define some country.
```

pedrobaeza (approved):

PR #246, opened 2021-12-01
[FIX] account_avatax: Reset to Draft button fix

atchuthan:
To simulate this error:

1. Post an invoice and transaction will be created to Avatax.
2. Lock the entries.
3. Click on Reset to draft button, and you will receive error "You cannot add/modify entries prior to and inclusive of the lock date" in odoo, but the transaction will be moved to voided state in Avatax.

So, I have changed the code such that we check all the Odoo conditions prior to sending the Avatax API call to void the transaction to avatax.

cc: @SodexisTeam @dreispt

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -336,6 +336,7 @@ class AccountMove(models.Model):
         """
         Sets invoice to Draft, either from the Posted or Cancelled states
         """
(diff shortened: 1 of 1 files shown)


dreispt (approved):

atchuthan:
cc: @bodedra @b-kannan

PR #250, opened 2021-12-15
[FIX] account_avatax_sale : compute sales tax each time sale order am…

### LINKED ISSUE #249: 13.0 account_avatax_sale wrong computation of sales tax when updating order
## Module
account_avatax_sale

## Describe the bug
Sales Tax computation using Avatax is wrong when updating sale order lines.

## To Reproduce
**13.0**:

1. Create new quotation for US customer with Sales Tax (for example "Space Needle - 400 Broad St, Seattle, WA 98109, United States")
2. Add a line with unit price = $1000
3. Click on Avalara "Compute tax" button : Untaxed amount = $1000, Tax = $102.5, Total amount = $1102.5
4. Save
5. Edit
6. Change unit price from $1000 to $2000 : Untaxed amount = $2000, Tax = $205, Total amount = $2205
7. Save, Untaxed amount = $2000, Tax = $102.5, Total amount = $2102.5
8. Edit
9. Remove the line, on screen we can see Untaxed amount = $0, Tax = $0, Total amount = $0
10. Save, Untaxed amount = $0, but Tax = $205 and Total amount = $205
11. Print Quotation / Order, in PDF file : Untaxed amount = $0, Tax = $0 (not displayed), but Total amount = $205

**Expected behavior**
A call to Avatax API should happen each time amounts are computed.

**Additional context**
A video demonstrating the issue (with erratic behaviour due to disconnection on runbot) : https://watch.screencastify.com/v/mBMHvCZP4SZW10AGcaF0

metaminux:
…ounts are computed.

Fixes #249

### DIFF
--- a/account_avatax_sale/models/sale_order.py
+++ b/account_avatax_sale/models/sale_order.py
@@ -78,7 +78,10 @@ class SaleOrder(models.Model):
         """
         super()._amount_all()
         for order in self:
-            if order.tax_amount:
+            if not order.order_line:
+                order.onchange_reset_avatax_amount()
+            elif order.tax_amount:
+                order.avalara_compute_taxes()
                 order.update(
                     {
                         "amount_tax": order.tax_amount,


account_avatax_sale/models/sale_order.py
```diff
@@ -78,7 +78,10 @@ def _amount_all(self):
         """
         super()._amount_all()
         for order in self:
-            if order.tax_amount:
+            if not order.order_line:
+                order.onchange_reset_avatax_amount()
+            elif order.tax_amount:
+                order.avalara_compute_taxes()
```
dreispt, line 84: @metaminux This will trigger an API call on each computation.
You should not do this, it is against Avalara's policy for clients.
There is risk to get unexpected bills for the service or have requests rate limited and blocked.

metaminux: @dreispt Thank you for this review. Do you have any idea on how to fix this behaviour in another way so that payment amount is right when sending orders to customers ?
Besides, I can't find any documentation on rate limit for Avalara : https://developercommunity.avalara.com/s/question/0D55a00008028Z9CAI/avatax-api-rate-limit but I agree to do it another way if amounts are right at the end.

dreispt: @metaminux At That's a fair use case

dreispt: @metaminux For [certification requirements](https://developer.avalara.com/erp-integration-guide/sales-tax-badge/functional-review/): 

> The quote-to-cash process should specifically call out the times when the system is reaching out to AvaTax for a tax calculation and when the system is going to record a transaction to Avatax for further reporting and remittance.

 and 

> Show that a reasonable API request ratio is used during invoicing.
> Sales invoices or requests that use document type SalesInvoice should not call AvaTax multiple times for the same invoice.

---

You are describing a reasonable scenario.
Maybe we need some "Needs Tax Recomputation" dirty flag, and prevent "Send Mail" button or print "Sale Order" report on that case?

PR #252, opened 2022-01-03
[15.0][MIG]account_fiscal_position_partner_type

manuelregidor:
Standard migration.

PR #253, opened 2022-01-04
[15.0][MIG]l10n_eu_oss

manuelregidor:
Standard migration
Depends on https://github.com/OCA/account-fiscal-rule/pull/252


ValentinVinagre:
@manuelregidor could you make a rebase please?

ValentinVinagre (approved):
LGTM 👍🏻

ValentinVinagre:
@HaraldPanten

HaraldPanten (approved):
Wizard is working fine for creating taxes and fiscal positions as well.

LGTM

emagdalenaC2i:
@manuelregidor Odoo 15 already has a module called l10n_eu_oss.
Won't this create some issues?

HaraldPanten:
That's right, seems that they made changes to their original module for OSS.

Maybe we should rename the module to l10n_eu_oss_oca

@pedrobaeza @ValentinVinagre WDYT?

emagdalenaC2i:
I did not check the Odoo module, I have just seen that it is called the same

HaraldPanten:
It seems that they improved their old system for applying OSS taxes, but I think they still not check the partner type (company/final customer) so... I still don't see an easy way to map the taxes according the customer type.

On the other hand, I'm not sure if they mapped the taxes correctly. For example, they do not consider the super reduced VAT for Spain. It's not the most common for distance sales, maybe, but it does exist.

IMO, we still need this module for V15 in OCA. So... renaming is the best option I think.

ValentinVinagre:
> It seems that they improved their old system for applying OSS taxes, but I think they still not check the partner type (company/final customer) so... I still don't see an easy way to map the taxes according the customer type.
> 
> On the other hand, I'm not sure if they mapped the taxes correctly. For example, they do not consider the super reduced VAT for Spain. It's not the most common for distance sales, maybe, but it does exist.
> 
> IMO, we still need this module for V15 in OCA. So... renaming is the best option I think.

I agree with @HaraldPanten

pedrobaeza:
But if the name is exactly the same, we should rename it.

ValentinVinagre:
> But if the name is exactly the same, we should rename it.

How should we make the name change? With an imp? the name could be: l10n_eu_oss_oca

pedrobaeza:
It's too late for doing things OK, but let's rename it on a new PR, and on v16, we need to extract whole commit history and do the renaming backwards.

PR #261, opened 2022-01-24
[FIX]l10n_eu_oss_oca: Prevent errors when creating invoices using OSS FP

manuelregidor:
Steps to reproduce the error:

1. A new invoice is created and an OSS Fiscal Postion is set (e.g. 'Intra-EU B2C in France (EU-OSS-FR)')
l10n_eu_oss_oca_1
2. When creating a new line, the right tax is set on the invoice line.
l10n_eu_oss_oca_2
3. If we try to save the invoice, an error occurs that indicates that we are trying to use taxes from an unallowed country
l10n_eu_oss_oca_3

That happens because there is a new computed field named tax_country_id in model account.move that retricts which taxes can be used depending on the country. In the example, because the Fiscal Postion 'Intra-EU B2C in France (EU-OSS-FR)' has France set as the country, if we do not have a VAT number established in that Fiscal Position in field foreign_vat, the  field tax_country_id in the invoice is not set to France so we cannot use a tax that is applicable to France.

The proposed solution adds a field to the fiscal position model to indicate whether a fiscal position belongs to the OSS system. This field is considered when the field tax_country_id in invoices is calculated, so the country established in the invoice is the one set in the fiscal position.

### DIFF
--- a/l10n_eu_oss_oca/__manifest__.py
+++ b/l10n_eu_oss_oca/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "L10n EU OSS OCA",
-    "version": "15.0.2.0.0",
+    "version": "15.0.2.1.0",
     "category": "Accounting & Finance",
     "website": "https://github.com/OCA/account-fiscal-rule",
     "author": "Sygel Technology," "Odoo Community Association (OCA)",
@@ -17,5 +17,6 @@
         "data/oss.tax.rate.csv",
         "wizard/l10n_eu_oss_wizard.xml",
         "views/res_config_settings.xml",
+        "views/account_fiscal_position_views.xml",
     ],
 }
--- a/l10n_eu_oss_oca/migrations/15.0.2.1.0/post-migration.py
+++ b/l10n_eu_oss_oca/migrations/15.0.2.1.0/post-migration.py
@@ -1,0 +1,19 @@
+# Copyright 2021 Sygel - Manuel Regidor
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        UPDATE account_fiscal_position afp
+        SET oss_oca = True
+        WHERE afp.country_id IS NOT NULL AND
+        afp.vat_required IS False AND
+        afp.auto_apply IS True AND
+        afp.fiscal_position_type = 'b2c'
+        """,
+    )
--- a/l10n_eu_oss_oca/models/__init__.py
+++ b/l10n_eu_oss_oca/models/__init__.py
@@ -3,3 +3,5 @@
 
 from . import oss_tax_rate
 from . import account_tax
+from . import account_move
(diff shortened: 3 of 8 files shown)


HaraldPanten:
@pedrobaeza @dreispt Due to new fields in V15, OSS OCA module needs to be updated.

Could you have a look?

THX

l10n_eu_oss_oca/__manifest__.py
```diff
@@ -3,19 +3,20 @@
...
+    "version": "15.0.2.1.0",
     "category": "Accounting & Finance",
     "website": "https://github.com/OCA/account-fiscal-rule",
     "author": "Sygel Technology," "Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "application": False,
     "installable": True,
     "development_status": "Production/Stable",
-    "depends": ["account", "account_fiscal_position_partner_type"],
+    "depends": ["account", "account_fiscal_position_partner_type", "sale"],
```
pedrobaeza: This new dependency is not OK. You can perform accounting without sales.

ValentinVinagre: We added the sales dependency because the same fix would also apply to sales. Do you want us to create a new module for the extension in sales?

pedrobaeza: Yes. Always respect modularity.

pedrobaeza:
I don't get the problem by the given description anyway. Remember to give steps to reproduce the problem or a better explanation.

HaraldPanten:
THX @manuelregidor for your explanation.

Pedro, Daniel, the steps to reproduce the problem have been updated 👍

HaraldPanten (approved):
LGTM

pedrobaeza:
Without this patch, the solution is to put the country and the EU VAT at the partner, isn't it? And isn't that the goal of this check and should be kept?

manuelregidor:
@pedrobaeza Without the patch, we need a valid foreign VAT at the fiscal position that allows us to operate in the country where the fiscal position applies. Otherwise, we cannot use OSS taxes in invoices with an OSS fiscal position, as the country where the mapped taxes apply and the value of tax_country_id in the invoice do not match.

pedrobaeza:
OK, understood. And I wonder, has this happened due to reusing the same field that exists? In previous versions, IIRC we created a field `oss_country_id` for storing this. Can we restore it?

manuelregidor:
@pedrobaeza The field oss_country_id is still set when the OSS taxes are created by the wizard in this module. We have thought of two solutions.
1. The first option is the one in this PR. 
2. The second option is setting our company's country as the country in OSS taxes when they are created by the wizard. In this case, the field tax_country_id in invoices and the field country_id in the taxes applied on the invoice lines will match, as they both will have the company's country.

We think that the solution we have proposed is more suitable, as it might be a bit confusing if a tax called, for instance, 'OSS for EU to Croatia: 25.0', has a country different from Croatia set as the tax country. What do you think?

pedrobaeza (approved):
OK, let's go with this solution

HaraldPanten:
Do you know if ocabot merge porcesses got "frozen" or something? @sbidoul

pedrobaeza:
It seems so. Just retry now.

HaraldPanten:
Let's see... THX Pedro

PR #262, opened 2022-01-24
[10.0][MIG]l10n_eu_oss

manuelregidor:
Backport to v10.


manuelregidor:
@HaraldPanten @ValentinVinagre

HaraldPanten (approved):
I could reproduce some "easy path" use cases in runboat, and it's working fine for me.

@manuelregidor could you change your commit message? Normally we should put [MIG] instead of [BCK]. Shouldn't we, @OCA/accounting-maintainers ?

HaraldPanten:
Fast tracking, as we are (probably) the only partner supporting that for V10 and all checks are green.

We'll be using that in a production environment, so... we'll take care if there are some issues.

PR #264, opened 2022-02-10
[13.0][FIX] account_avatax: use active company in multicompany settings

dreispt:


atchuthan (approved):
:+1:

smangukiya (approved):

smangukiya:
@dreispt It seems we need to take care of code coverage %

PR #265, opened 2022-02-24
[FIX] account_avatax: Void transaction from posted status

atchuthan:
There was an error with sending void transaction and corrected it now.

cc: @SodexisTeam @dreispt


dreispt:
Thanks, can you fix pre-commit please?

atchuthan:
> Thanks, can you fix pre-commit please?

Done pre-commit

dreispt (approved):
The change seems functionally equivalent to the old code, so I'm not sure what is being fixed exactly. But I see no reason to not merge.

PR #270, opened 2022-03-04
[14.0][FIX] account_product_fiscal_classification: apply taxes if a default classification is specified using ir.default

kevinkhao:


account_product_fiscal_classification/models/product_template.py
```diff
@@ -112,24 +112,31 @@ def fields_view_get(
                 result["arch"] = etree.tostring(doc)
         return result
 
+    def _apply_classification_taxes(self):
+        # update or replace 'taxes_id' and 'supplier_taxes_id'
+        classification = self.fiscal_classification_id
+        tax_vals = {
+            "supplier_taxes_id": [
+                (6, 0, [x.id for x in classification.sudo().purchase_tax_ids])
```
dreispt: Simpler like this?
```suggestion
                (6, 0, classification.sudo().purchase_tax_ids.ids)
```

account_product_fiscal_classification/models/product_template.py
```diff
@@ -112,24 +112,31 @@ def fields_view_get(
...
+            ],
+            "taxes_id": [(6, 0, [x.id for x in classification.sudo().sale_tax_ids])],
```
dreispt:
```suggestion
            "taxes_id": [(6, 0, classification.sudo().sale_tax_ids.ids)],
```

account_product_fiscal_classification/models/product_template.py
kevinkhao: indeed, thanks

dreispt:
@legalsylvain You're the maintainer here. Can you review please?

legalsylvain:
Hi @dreispt : thanks for pinging me. a pity that oca bot doesn't ping me. the feature was working before, AFAIK, right ?

@kevinkhao : I don't understand your patch. when ir.default is defined, the values should be in the vals, in the ``create(self, vals)`` part. so it should work without that patch ? 
or did I missed something ?

EDIT : 
how I see...
```
            # special case: using ir.default for fiscal_classification_id means
            #   value is not in vals, but is actually present on record
```
weird corner case but oK.


Just a question : I don't understand the name of ``create_mode_fiscal_sync``. what does ``sync`` means ?
something like ``create_mode_fiscal_classification`` seems better. (or did I missed something)...

PR #273, opened 2022-03-10
[14.0][IMP] account_fiscal_position_partner_type: FP with empty FP Type should apply to any Partner type

### LINKED ISSUE #254: account_fiscal_position_partner_type: suprising behavior on Fiscal Positions
I would expect that a Fiscal Position with empty Fiscal Position Type would apply to any Partner.
I should be able to create a single FP "National Customers", that applies both to B2B and B2C.

Right now this doesn't work.
I need to instead have at least two FPs, duplicates except for the FP Type field: one for B2C and another for B2C.

I would like to change this, and allow an FP with no FP Type set to be applicable to Partners with any FP Type assigned to the,

Thoughts? @pedrobaeza @HaraldPanten @ValentinVinagre

dreispt:
Closes https://github.com/OCA/account-fiscal-rule/issues/254

### DIFF
--- a/account_fiscal_position_partner_type/models/account_fiscal_position.py
+++ b/account_fiscal_position_partner_type/models/account_fiscal_position.py
@@ -16,19 +16,12 @@ class AccountFiscalPosition(models.Model):
 
     @api.model
     def search(self, args, offset=0, limit=None, order=None, count=False):
-        if self.env.context.get("fiscal_position_type"):
-            args = expression.AND(
-                (
-                    args,
-                    [
-                        (
-                            "fiscal_position_type",
-                            "=",
-                            self.env.context["fiscal_position_type"],
-                        )
-                    ],
-                )
-            )
+        add_domain = [
+            "|",
+            ("fiscal_position_type", "=", self.env.context["fiscal_position_type"]),
+            ("fiscal_position_type", "=", False),
+        ]
+        args = expression.AND((args, add_domain))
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
I think this is not needed. If you don't fill the fiscal position type, no context is passed, so the query expression is not modified and no filter is done.

dreispt:
@pedrobaeza I just revisited the discussion on this:

> If I don't remember bad, not filling the type on the fiscal position means to not check the type of the partner, so that sounds exactly what you want to achieve. You haven't indicated the Odoo version though.

Right now the above is not true - the Partner Type is still checked if unset on the FP.
This change allows for the Partner Type to not be checked when not set in the FP.

Am I misunderstanding this?

pedrobaeza:
The solution to your problem is to set a default value for `fiscal_position_type`. Both design decisions (current one and the one you propose) has its flaws and advantages, but the current one was chosen for maximizing the compatibility: if nothing is selected, nothing from standard is changed. cc @HaraldPanten

HaraldPanten:
> The solution to your problem is to set a default value for `fiscal_position_type`. Both design decisions (current one and the one you propose) has its flaws and advantages, but the current one was chosen for maximizing the compatibility: if nothing is selected, nothing from standard is changed. cc @HaraldPanten

If I remember well, it works as you say.

PR #274, opened 2022-03-16
[14.0] [FIX] Improved the account_avatax module code.

Chandresh-OSI:


dreispt (approved):
Did functional tests and works as expected.

metaminux:
Hi @Chandresh-OSI and @dreispt 

Sorry to put a comment here 6 months after merging, but with this commit, there is no more check on the line price_unit, so having 2 lines with same product and quantity but different prices is messing the line lookup...

Do I miss something here ? Why removing these lines at first point ?

I would like some explanation before creating a new issue with clear example.

Thanks,

dreispt:
@metaminux Don't be sorry, that's part of continuous improvement.
I see your point, that case wasn't verified (and we kind of miss automated tests).
Can you submit a fix please?

PR #285, opened 2022-04-28
[FIX][14.0] account_avatax

Vandan-OSI:
Fixed AVATAX Validation  Error while Adding Line/Section at Invoice Line.

image

**Reproduce Steps:**
1). Create Sale Order with the Non-Validate Address (Change the address of the Sales Order after Confirming the Address).
2). Now Create an Invoice for that Sale Order.
3). Now Add the Section or  Note in the Invoice it will Trigger a Validation Error.

**Desired Output:**
When we Add the Section or  Note in the Invoice we are not Doing any calculation.

### DIFF
--- a/account_avatax/models/account_move.py
+++ b/account_avatax/models/account_move.py
@@ -380,7 +380,7 @@ class AccountMove(models.Model):
                     line._origin.price_unit != line.price_unit
                     or line._origin.discount != line.discount
                     or line._origin.quantity != line.quantity
-                ):
+                ) and not line.display_type:
                     self.calculate_tax_on_save = True
                     break
 


account_avatax/models/account_move.py
```diff
@@ -380,7 +380,7 @@ def onchange_avatax_calculation(self):
                     line._origin.price_unit != line.price_unit
                     or line._origin.discount != line.discount
                     or line._origin.quantity != line.quantity
-                ):
+                ) and line.display_type not in ("line_section", "line_note"):
```
dreispt: Usually we check if display_type is set
```suggestion
                ) and not line.display_type
```

dreispt (commented):
Thank you @Vandan-OSI, please squash the commits, so we can forward port this for v15.
Also remove the "[14.0]" from the commit message.
Thanks.

atchuthan (approved):

dreispt (approved):

PR #287, opened 2022-05-18
[ADD][14.0]account_avatax_sale_fire_protection

Vandan-OSI:
This module allows For Delivery Addresses in the state of Florida and applies conditional replacement on Sale Order Products of non-taxable Labor Products.


Vandan-OSI:
@dreispt  Can you Please Review this PR

account_avatax_sale_fire_protection/models/sale_order.py
```diff
@@ -0,0 +1,126 @@
...
+
+    is_fl_taxation = fields.Boolean(
+        string="Florida Labor Taxable", compute="_compute_is_fl_taxation"
+    )
+    is_fl_address = fields.Boolean(
+        string="Florida Address", compute="_compute_is_fl_address"
+    )
+
+    def _get_service_product_line(self):
+        for rec in self:
```
atchuthan: You could use self.ensure_one() and directly return value

```
self.ensure_one()
return self.order_line.filtered(lambda p: p.product_id.type == "service")
```

Vandan-OSI:
@dreispt Can you please reopen this PR?

CC @atchuthan

PR #290, opened 2022-06-23
[FIX] account_avatax: Field 'salespersonCode' has an invalid length

atchuthan:
Field 'salespersonCode' must be between 0 and 25 characters in length

Screen Shot 2022-06-20 at 12 32 34 PM

cc: @SodexisTeam @dreispt


atchuthan:
@dreispt Please review

dreispt (approved):

dreispt:
Thanks @atchuthan 
It would be nice to forward port this to v15

PR #294, opened 2022-07-07
[14.0] account_avatax: Add Avatax retail price

atchuthan:
cc: @SodexisTeam @dreispt

### DIFF
--- a/account_avatax_oca/__manifest__.py
+++ b/account_avatax_oca/__manifest__.py
@@ -24,6 +24,7 @@
         "views/account_move_view.xml",
         "views/account_tax_view.xml",
         "views/account_fiscal_position_view.xml",
+        "views/retail_group.xml",
     ],
     "images": ["static/description/avatax_icon.png"],
     "installable": True,
--- a/account_avatax_oca/models/__init__.py
+++ b/account_avatax_oca/models/__init__.py
@@ -5,4 +5,5 @@ from . import account_move
 from . import account_fiscal_position
 from . import account_tax
 from . import res_company
+from . import retail_group
 from . import avatax_rest_api
--- a/account_avatax_oca/models/account_move.py
+++ b/account_avatax_oca/models/account_move.py
@@ -266,12 +266,30 @@ class AccountMove(models.Model):
             for index, line in enumerate(lines):
                 tax_result_line = tax_result_lines.get(line.id)
                 if tax_result_line:
-                    rate = tax_result_line.get("rate", 0.0)
-                    tax = Tax.get_avalara_tax(rate, doc_type)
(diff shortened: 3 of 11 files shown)


dreispt:
Thank you @atchuthan 
I have concerns about the design for this feature:
I wonder if it is duplicating some Odoo features, and I'm not sure how it is expected to be used.

It would be helpful the have an explanation on how to use, and the how it works.
Ideally, you would add this into the proper readme fragments.

Thanks!

dnplkndll:
@dreispt  would our next CO filing be useful to review? it just took effect this month so that will be after 8/1/22. all the US businesses with CO nexus would have this liability but might not be any others using Odoo?

atchuthan:
> Thank you @atchuthan I have concerns about the design for this feature: I wonder if it is duplicating some Odoo features, and I'm not sure how it is expected to be used.
> 
> It would be helpful the have an explanation on how to use, and the how it works. Ideally, you would add this into the proper readme fragments.
> 
> Thanks!

Sure, we will add the info in readme fragments and publish the PR later

osi-scampbell:
@atchuthan is there an update on this PR?

atchuthan:
> @atchuthan is there an update on this PR?

We are currently using this fix for one of our customers in production.

atchuthan:
> Thank you @atchuthan I have concerns about the design for this feature: I wonder if it is duplicating some Odoo features, and I'm not sure how it is expected to be used.
> 
> It would be helpful the have an explanation on how to use, and the how it works. Ideally, you would add this into the proper readme fragments.
> 
> Thanks!

@dreispt I don't think this is some duplicate of odoo features.  We are adding a new line based on config to SO line and Invoice line.

Sure, we will add the explanation in readme fragments in few days.

cybernexus:
I'm trying to determine how this pull request works to satisfy the conditions set forth by Avalara? According to their information you must add a line item to sale / invoice for a product setup with the appropriate tax code and make the price for that line a $0.00 amount. I tried this in one of our staging environments. using this pull request, and it did not appear to work. As I read it, the line item should be added to sale / invoice and then multiple taxes should be applied. One a fixed tax of $0.27 and then others as well because you can have a tax on that $0.27 fee. I did some research and apparently this is not specific to the Colorado delivery fee. The avatax API allows for the ability to return mixed mode flat and percentage based tax rates in their response for any particular item line.    This should be patched as soon as possible because the side affect as of right now is that you could have a $1,000.00 line subject to a $20.00 flat rate fee but the avatax module sees it as a $200.00 fee. All tax rates get imported and converted to percentage regardless of the information stored in the "unitofBasis" field.

PR #301, opened 2022-07-19
[13.0][IMP] Backport changes from 14.0

metaminux:
Hi,

I took every commit from 14.0, cheery-picked and adapt to v13.

Main differences are about :
- type field for account.move model which is move_type in 14.0
- with_context(force_company=xx) which is with_company(xx) in 14.0
- method post(self) for model account.move is now _post(self, soft=True) in 14.0


dreispt:
It is curious that your cherry-picks don't include the original author information.
This would be an issue if this was a forward-port, but not a big concern for backport to an old Odoo version.
But you should double check the process you used to cherry-pick commits.

metaminux:
Hi @dreispt 

Yes, I used SmartGit for cherry-picking and I didn't know that not committing directly is not preserving the original author.
Don't know if I can modify that now... sorry.

I didn't commit directly to review every line for adaptation to 13.0.

Issue #304, opened 2022-07-20
[13.0][14.0] account_avatax,account_avatax_sale : wrong calculation for sales taxes with small amounts


metaminux:
## Module
- account_avatax
- account_avatax_sale

## Describe the bug
When dealing with small amounts, sales taxes are rounded and Avatax is sending different tax calculation for the same original price.
This gives strange but accurate amounts in sale order lines
Then, when creating the invoice, first amounts computed by Odoo are wrong.
Using the Compute tax button or posting the invoice is giving wrong amounts on lines, wrong total amount due and wrong journal entries
Finally, amounts declared to Avalara and amounts in Odoo are different, and documents issued to customers lead to wrong payment.

## To Reproduce

- Odoo v13
- account_avatax v13.0.3.4.4
- account_avatax_sale v13.0.2.4.4

1. Install a brand new Odoo v13 database with demo data
2. Install account_avatax and account_avatax_sale
3. Configure Avatax API with tax calculation enabled
4. Configure **Lines Subtotals Tax Display** as **Tax-Included** in Accounting settings
5. Create a customer from Denver, CO
6. Validate the address through Avatax
7. Create a sale order for this customer with 10 lines of the same product (for example DESK005, qty = 1) at $0.45 for unit price
8. Confirm the SO
9. Here, tax amounts per line are different, but comes from Avalara (see confirmed SO and Avalara transaction screenshots below)
10. Deliver all goods to be able to invoice the SO
11. Create the invoice
12. Here, tax amount computed by Odoo are wrong (see Draft invoice screenshot below)
13. Post the invoice
14. Here, tax amount are wrong per line and total amount due is wrong too (see Posted invoice and Journal entries screenshots below)
15. Journal entries are also wrong
16. Finally, when registering payment, the wrong amount due (see Register payment screenshot below)

**Expected behavior**

- Subtotal amount tax-included in each invoice line should be the one returned by Avalara
- Journal entry amount for total tax should the one returned by Avalara

### Confirmed SO
![2022-07-20_confirmed_SO](https://user-images.githubusercontent.com/9568711/180035476-13207320-73d5-4aa7-a1b1-35d8e33b889d.png)

### Avalara transaction
![2022-07-20_avalara_details](https://user-images.githubusercontent.com/9568711/180036900-ccfe0664-4f42-422d-85d1-7a1887c63ad1.png)

### Draft Invoice
[...]

PR #313, opened 2022-08-02
[FIX] account_avatax: Tax rounding not applied in Invoice

atchuthan:
Steps to reproduce
- Let's assume that a SO was created, and we received tax as $15.14 and 9.25% for $163.75(which should be $15.146875 ~ 15.15 as per calculation) but we take the tax from avatax as it is and set it in SO.
- Create an invoice and compute tax with Avatax with tax as $15.15(tax calculated based on odoo) and will have $0.1 residual amount difference
- Recompute tax once again, and the tax amount changes to $15.14(Tax amount from Avatax).

Our fix is to send the context when the taxes are added to the Invoice line.

cc: @SodexisTeam @dreispt

### DIFF
--- a/account_avatax_oca/models/account_move.py
+++ b/account_avatax_oca/models/account_move.py
@@ -281,7 +281,9 @@ class AccountMove(models.Model):
 
             # Set Taxes on lines in a way that properly triggers onchanges
             # This same approach is also used by the official account_taxcloud connector
-            with Form(self) as move_form:
+            with Form(
+                self.with_context(avatax_invoice=self, check_move_validity=False)
+            ) as move_form:
                 for index, taxes in taxes_to_set:
                     with move_form.invoice_line_ids.edit(index) as line_form:
                         line_form.tax_ids.clear()


metaminux:
Hi @atchuthan 

I'm trying to see if #304 is related but I've trouble reproducing your issue.

Assuming your Avalara advanced settings in On for Control Rounding.
avalara_control_rounding

I can get the amount you're describing for a customer in Tenessee (9.25% for $163.75 gives me $15.14 sales taxes)
tenessee_sale_order

When creating the invoice, Odoo computes a $15.15 tax (not using the amount computed for SO, see #304).
tenessee_draft_invoice

But as soon as I click on Compute Tax or Post, amounts are retrieved from Avatax and there is no residual difference amount.
tenessee_draft_recomputed

So I can't see the effect of your commit.

Can you tell me where I'm wrong ?

dreispt:
Hi @atchuthan these rounding issues is also a concern for me.
Can you share a particular example we can try too: specific delivery address and amount.
Thank you

PR #314, opened 2022-08-02
[15.0][REF] account_avatax_oca: rename from account_avatax, avoiding Odoo EE collision

### LINKED ISSUE #283: [14.0][15.0] account_avatax: updating Enterprise code breaks the Avatax connector!
Odoo SA rolled out their own `account_avatax` and `account_avatax_sale` models as EE modules, for 14.0 and 15.0.
This hijacks the OCA module names, and breaks current installations.

**If you are using OCA Avatax modules, don't update your Odoo Enterprise code!**

Upstream discussion:
https://github.com/odoo/enterprise/pull/23535

dreispt:
This is needed since Odoo Enterprise now provides a different module with the same name,
and the name collision had to be resolved.

To migrate to the new OCA module, first install ``account_avatax_oca`` and
``account_avatax_sale_oca``, and then uninstall ``account_avatax``.

Closes #283


MiquelRForgeFlow (approved):
Could you do the same for v14?

account_avatax_oca/hooks.py
```diff
@@ -0,0 +1,15 @@
...
+
+def pre_init_hook(cr):
+    # Preserve key data when moving from account_avatax to account_avatax_oca
+    # The process is to forst install account_avatax_oca
+    # and then uninstall account_avatax
+    cr.execute(
+        """
+        UPDATE ir_model_data
+        SET module = 'account_avatax_oca'
+        WHERE name in ('avatax_fiscal_position_us', 'avatax')
```
MiquelRForgeFlow:
```suggestion
        WHERE name in ('avatax_fiscal_position_us', 'account_avatax')
```

MiquelRForgeFlow (changes requested):

account_avatax_oca/hooks.py
dreispt: Thanks for reviewing - fixed.

dreispt:
@MiquelRForgeFlow Yes, will do for v14 too.

MiquelRForgeFlow (approved):

Issue #318, opened 2022-08-31
[15.0] [ERROR] [account_avatax] MissingLine


Nikul-OSI:
## Module
account_avatax

## Describe the bug
When Create new invoice with invoice line that time thought User Error: MissingLine

## To Reproduce
**Affected versions**:

![OCA Avatax issue](https://user-images.githubusercontent.com/98144048/187620619-ae7e629f-7a5b-4d2b-9602-33f1f2010fab.gif)

**Expected behavior**
create invoice with successful Avatax API call/log.

PR #319, opened 2022-08-31
[15.0][FIX][account_avatax] Fixed MissingLine

Nikul-OSI:
Fixed Issue: https://github.com/OCA/account-fiscal-rule/issues/318


rafaelbn:
Ping @dreispt 😄 happy new year!

PR #320, opened 2022-09-08
[14.0][REF] account_avatax_oca: rename from account_avatax, avoiding Odoo EE collision 

dreispt:


JordiBForgeFlow:
@dreispt Changing a module name within one version can break a lot of installations... It's typically best to introduce the new module name when a new version is launched. What is the problem with this name in v14?

dreispt:
@JordiBForgeFlow Indeed - please read https://github.com/OCA/account-fiscal-rule/issues/283

JordiBForgeFlow:
@dreispt I understand. Some installations will need to undergo this renaming of the module. Now they use OCA's account_avatax and now a migration process should make sure that they are renamed to account_avatax_oca. @MiquelRForgeFlow thoughts?

dreispt:
A manual process was provided.
If we can have a migration script to automate it, that would be great.

JordiBForgeFlow:
@dreispt the instructions that you refer to are https://github.com/OCA/account-fiscal-rule/pull/320/files#diff-3d8de764451586be84b7f33bc693a6f1a46a02178000b8e70a227656b6f5362fR7, right?

dreispt:
@JordiBForgeFlow Yes, and readme/DESCRIPTION.rst too.

JordiBForgeFlow:
@dreispt we'll review the PR and test the procedure then. Thanks!

MiquelRForgeFlow (approved):

PR #327, opened 2022-10-10
[16.0][MIG] account_product_fiscal_classification (from 14.0 to 16.0)

legalsylvain:
Migration of ``account_product_fiscal_classification`` from 14.0 to 16.0.
- remove ``account_product_fiscal_classification_test``. (now all tests are done in the same module).
- increase coverage of the module. 81,9% -> 86,5%
- code simplification. (python lines : 526 -> 444) = -15%
- Remove ``usage_group_id`` field on classification and fiscal classification on product.category, replace by generic new model ``account.product.fiscal.rule``.
- Improve installation of the module. (now install module on huge database takes second, before minutes).
- remove double write access on DB when creating or writing ``product.template`` models.
- improve and update documentation.

### Description

This module extends the Odoo Account module to introduce Fiscal Classifications.

A fiscal classification is an association of purchase and sale taxes that have to be set on your products.

product_template_form

This new design provides the following improvments.

* It is more simple (less clicks quantity) to configure correctly a product. (You only have to set a fiscal classification, instead of setting one or many sale taxes, then click on purchase tab and set one or many purchase taxes.)

* Prevent users to select incompatible purchase and sale taxes. French Exemple: A product can not be configured with:

  * Purchase Taxes: 5.5%
  * Sale Taxes: 20%

* Optionaly, you can restrict the usage of some fiscal classifications to some product categories.
  French Exemple : A product set in the category "Beer and wine" can not have a sale tax 5,5%. (20% is mandatory)

### Configure

* Go to "Accounting > Configuration > Accounting > Fiscal Classifications".

* Create new classification, with correct purchase and sale taxes.

fiscal_classification_form

**Note**

When installing the module, it will create fiscal classifications for all the taxes combinations that exist in your database.
If some are bad, you can massively move the products for a bad classification to the good one, using the button "Move products into another fiscal classification".

**Optional settings**

Once the fiscal classifications are created, you can create rules to avoid users to use some classifications that don't make sense for some categories.

* Go to "Accounting > Configuration > Accounting > Fiscal Classification Rules"
* Create new rules.

fiscal_classification_rule__wine

Every time a new product is created, the rules that matches the product setting will be checked **in the displayed order**.

* If one rule is set to 'Allow', the next rules are ignored.
* If one rule is set to 'Forbid', an error is raised.

fiscal_classification_rule__wine_warning


alexis-via:
Great to see the end of the double write... i guess we can start to use this module now :)

I would suggest the use of "_check_company_auto = True" on account.product.fiscal.classification. I'll make a PR.

legalsylvain:
> Great to see the end of the double write... i guess we can start to use this module now :)

Nice ! 

> I would suggest the use of "_check_company_auto = True" on account.product.fiscal.classification. I'll make a PR.

you can do it against legalsylvain:16.0-mig-account_product_fiscal_classification, I'll merge it quickly.

alexis-via:
If we use _check_company_auto = True, it would be more appropriate on product.template, not on account.product.fiscal.classification. There is no _check_company_auto by default on product.template, so...
BTW, I think we should align the behavior of the default value for company_id of account.product.fiscal.classification with the behavior on product.template. It seems to me that, by default, company_id is False on product.template (it is a configuration param somewhere ?) ; so maybe should also have company_id = False by default on account.product.fiscal.classification, don't you think ?

rafaelbn:
Thank you!

@moduon MT-1618

account_product_fiscal_classification/__manifest__.py
```diff
@@ -0,0 +1,39 @@
...
+{
+    "name": "Account Product - Fiscal Classification",
+    "summary": "Simplify taxes management for products",
+    "version": "16.0.1.0.1",
+    "category": "Accounting",
+    "author": "Akretion,GRAP,La Louve,Odoo Community Association (OCA)",
+    "maintainers": ["legalsylvain"],
+    "website": "https://github.com/OCA/account-fiscal-rule",
+    "license": "AGPL-3",
+    "depends": ["account", "account_usability"],
```
rafaelbn, line 15: Is it required to depends in `account_usability`? I understand `account_usability `but I don't use this module and I would like to maintain without using it.

Just a question. Thank you! 😄

rafaelbn (commented):
We have the need to set fiscal positions to product that over write the fiscal position of the partner

- Product fiscal position has more priority than partner fiscal position

But not always, only on some specific products

account_product_fiscal_classification/__manifest__.py
legalsylvain: Hi @rafaelbn. account_usability module (previously named account_menu module in v<16) is a simple module that adds menu entries. It is required to put the menu of the classification template in the correct place. IMO : 

- the module is well maintained. (present in all the version)
- it is easy (mostly xml)

rvalyi: > Hi @rafaelbn. account_usability module (previously named account_menu module in v<16) is a simple module that adds menu entries. It is required to put the menu of the classification template in the correct place. IMO : 
> 
> - the module is well maintained. (present in all the version)
> - it is easy (mostly xml)

Hello, Sylvain, I have the same doubt for other modules I want to port to v16. One of the doubts I have is if people will complain if it doesn't play well with Odoo EE. Do you know if account_menu works decently along with Odoo EE?

legalsylvain:
Hi @rafaelbn 

> We have the need to set fiscal positions to product that over write the fiscal position of the partner

>    Product fiscal position has more priority than partner fiscal position

> But not always, only on some specific products

* I'm not sure to understand your need, could you elaborate with a real exemple
* At the first sight, adding fiscal position on product should be done in another module. The objective of this one is only to improve fiscal taxes settings on product.

@alexis-via 

> It seems to me that, by default, company_id is False on product.template (it is a configuration param somewhere ?) ; so maybe should also have company_id = False by default on account.product.fiscal.classification, don't you think ?

hum... 
- in V12 (and before) : default company of product template is the current one. ([See](https://github.com/odoo/odoo/blob/12.0/addons/product/models/product_template.py#L110))
- in V13 (and after) : company of product template is False by default. ([See](https://github.com/odoo/odoo/blob/13.0/addons/product/models/product_template.py#L121))

I don't like the new design, because that's not the one I implemented in my multicompany context !
But well, you're probably right, we should have a consistent logic.
I removed the default on account.product.fiscal.classification.

@rafaelbn, @alexis-via : could you update your review ? 

thanks, and happy new years !

account_product_fiscal_classification/__manifest__.py
legalsylvain: Hi. 

> Do you know if account_menu works decently along with Odoo EE?

I don't understand why you ask me ? ;-) 
I don't have access to the EE enterprise code, and your akrétion team has access.
So only you can say if there are incompability ! 

But technically, we have to understand that OCA modules are only "guaranteed" for the CE version, by design (there is no OCA-EE runboat, and no OCA-EE test).

rvalyi: yes, I understand. That being said lot's of OCA modules got little modifications to play it nice with Odoo EE as many OCA users also have customers with EE. As for me I don't have access to a v16 EE repo and that's not a big deal/blocker for me, just "a better have" in case you knew the answer about compatibility.

Shide (approved):
LGTM

rvalyi (approved):

rafaelbn:
@legalsylvain could you check failure please? Thank you!

PR #330, opened 2022-11-01
[ADD] account_default_fiscal_position

remytms:
Set a default fiscal position for partner when creating one. It uses the country referenced on the partner to link the right fiscal position.

Set the default fiscal positions for countries in
 - Accounting > Configuration > Accounting > Fiscal Positions.

If a partner is created for those set countries, the fiscal position will be set.


remytms:
I think that this module is not needed anymore in Odoo 14.0 as it can be done in standard.

Issue #344, opened 2023-01-05
14.0 account_product_fiscal_classification : Traceability on changes


bealdav:
**Describe the solution you'd like**
Because this module have a big impact on products and accounting
I would like add chatter info to `account.product.fiscal.classification` 
i.e. get info about updated tax m2m on this object, and name field
For m2m it would required to use `message_post` in `write` to get info about changes

**Describe alternatives you've considered**
Pray that anybody try things ! No always proof 

cc @legalsylvain @florian-dacosta @sebastienbeau

legalsylvain:
Hi. the track_visibility is enabled on product, [thanks to you](https://github.com/OCA/account-fiscal-rule/commit/463bb824941759c2472233c0ee80b16efbdd5115) ! (And yes, it's important).

What about to make the same, regarding the fields sale_tax_ids and purchase_tax_ids (and name) ? 

(and adding inheritance in the module account.product.fiscal.classification to have the chatter enabled ?)

bealdav:
Hi  @legalsylvain 
For m2m fields I plan to override `write()` and call `message_post()` inside to define old and new values

Ok for that ?

legalsylvain:
>  For m2m fields I plan to override write() and call message_post() inside to define old and new values

        track_visibility="onchange" doesn't work for m2m ?

legalsylvain:
Side question : could you review this one ? https://github.com/OCA/account-fiscal-rule/pull/327

bealdav:
Nope

>    track_visibility="onchange" doesn't work for m2m ?

legalsylvain:
>    track_visibility="onchange" doesn't work for m2m ?

[Damn'it !](https://i.giphy.com/media/OPQkiDUE3GSiRQacfa/giphy.webp)

So indeed, we should override write function. Could you make a PR against https://github.com/OCA/account-fiscal-rule/pull/327 ? 

Thanks !

bealdav:
I'll make a PR on 14.0 first and then apply changes on your PR 16.0, isn't it ?

legalsylvain:
> I'll make a PR on 14.0 first and then apply changes on your PR 16.0, isn't it ?

as you wish !

PR #345, opened 2023-01-05
[14.0] [IMP] acc_fisc_classif: add chatter on categ and log tax m2m

bealdav:
cc @legalsylvain could you evaluate it, please ?

For the futur I suppose we could use this one https://github.com/OCA/server-tools/pull/2434

But now we can merge as it, I suppose

### DIFF
--- a/account_product_fiscal_classification/models/account_product_fiscal_classification.py
+++ b/account_product_fiscal_classification/models/account_product_fiscal_classification.py
@@ -2,19 +2,22 @@
 # @author: Sylvain LE GAL (https://twitter.com/legalsylvain)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from collections import defaultdict
+
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
 
 class AccountProductFiscalClassification(models.Model):
     _name = "account.product.fiscal.classification"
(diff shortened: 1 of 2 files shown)


bealdav:
FIX https://github.com/OCA/account-fiscal-rule/issues/344

legalsylvain:
Tested on runboat. LGTM.

Tecnically, I was asking to me if it was not interesting to create a dedicated module name tracking_x2x with something generic. What do you think ? (can be done in a second step)

image

bealdav:
Yes you're right, it should be done in a separate helper module.
In a second step yes. We should go for v16 for that ?

LESTRAT21:
Hello, I tested via runboat. This module is suppose to add a chatter in the fiscal classification view, isn't it ? I cannot see the chatter.
image
 I did something wrong  ?

LESTRAT21:
I didn't tested non the right runboat. My functionnal test is ok : in the chatter, the taxes added in the the classification category are listed.
image

LESTRAT21 (approved):

sebastienbeau:
@legalsylvain @LESTRAT21 the module tracking_manager is ready for V14, already used in production (all m2m can be tracked)

legalsylvain:
@sebastienbeau : where is the module tracking_manager ?

PR #347, opened 2023-01-06
[16.0] [MIG] account_avatax_oca

Nikul-OSI:
* Added **sale_stock** as depends because of stock group already used in module. https://github.com/OCA/account-fiscal-rule/pull/347/files#diff-cbacfb36ea6c5fbb9accf7e15e9da58a871ebe4e0110eb82857c154ac78db972R40


Nikul-OSI:
@dreispt @atchuthan can you please take a look?

dreispt (approved):
This is here for a while, it is being used in projects, and we even have fix PRs for it.
So I think it is time to have it merged.

PR #353, opened 2023-03-16
[FIX] account_avatax_oca: Don't change the include_caba_tags argument

johnw-bluemark:
I'm not sure why `include_caba_tags` is forced to `False` here, but it looks like a mistake to me, and there's no comment explaining it. Passing it through unaltered allows the following enterprise tests to pass:

account_reports.tests.test_tax_report.TestTaxReport.test_caba_always_exigible

### DIFF
--- a/account_avatax_oca/models/account_tax.py
+++ b/account_avatax_oca/models/account_tax.py
@@ -77,7 +77,7 @@ class AccountTax(models.Model):
             partner,
             is_refund,
             handle_price_include,
-            include_caba_tags=False,
+            include_caba_tags,
         )
         avatax_invoice = self.env.context.get("avatax_invoice")
         if avatax_invoice:
--- a/account_avatax_sale_oca/models/account_tax.py
+++ b/account_avatax_sale_oca/models/account_tax.py
@@ -23,7 +23,7 @@ class AccountTax(models.Model):
             partner,
             is_refund,
             handle_price_include,
-            include_caba_tags=False,
+            include_caba_tags,
         )
         for_avatax_object = self.env.context.get("for_avatax_object")
         if for_avatax_object:


dreispt:
It is not being forced to False.
It is a default value if not provided.
If you remove it, the argument will them be a required one, and eventual calls missing it will break.

johnw-bluemark:
This is the function definition

```python
    def compute_all(
        self,
        price_unit,
        currency=None,
        quantity=1.0,
        product=None,
        partner=None,
        is_refund=False,
        handle_price_include=True,
        include_caba_tags=False,
    ):
```
If I removed the `=False` from here, then what you said would be sort of true, but it would actually just syntax error, as it is illegal to have positional arguments after keyword arguments. But that isn't what I changed. I changed the call to the original function.

```python
        res = super().compute_all(
            price_unit,
            currency,
            quantity,
            product,
            partner,
            is_refund,
            handle_price_include,
            include_caba_tags=False,
        )
```

As this is written, all of the arguments are passed positionally, except for the last one, which is a keyword argument with a value that is always False. That syntax only creates default arguments in function definitions, not function calls. You could fix it by writing it like this:

```python
        res = super().compute_all(
            price_unit,
            currency,
            quantity,
            product,
            partner,
            is_refund,
            handle_price_include,
            include_caba_tags=include_caba_tags,
        )
```

This will work correctly, but is a little verbose, so I wrote it like this:

```python
        res = super().compute_all(
            price_unit,
            currency,
            quantity,
            product,
            partner,
            is_refund,
            handle_price_include,
            include_caba_tags,
        )
```

If you doubt my words, run this script

```python
class A(object):
    def my_func(self, arg=False):
        return arg

class B(A):
    def my_func(self, arg=False):
        return super().my_func(arg=False)
        
class C(A):
    def my_func(self, arg=False):
        return super().my_func(arg=arg)

class D(A):
    def my_func(self, arg=False):
        return super().my_func(arg)

# create instances
b = B()
c = C()
d = D()


print(f"B: {b.my_func(True)}")
print(f"C: {c.my_func(True)}")
print(f"D: {d.my_func(True)}")
```

Output:
```
B: False
C: True
D: True
```

dreispt:
You are right.
I didn't expand the diff and for some reason assumed you were changing the function signature, not the super call, as you are.
Sorry to have you go through that long explanation.

dreispt (approved):

johnw-bluemark:
Is this waiting on something?

PR #361, opened 2023-05-15
[16.0] account_product_fiscal_classification: 2 small changes

alexis-via:
2 small changes:
1) add search view on account.product.fiscal.classification
2) remove translate=True on name field on account.product.fiscal.classification

### DIFF
--- a/account_product_fiscal_classification/models/account_product_fiscal_classification.py
+++ b/account_product_fiscal_classification/models/account_product_fiscal_classification.py
@@ -10,7 +10,7 @@ class AccountProductFiscalClassification(models.Model):
     _name = "account.product.fiscal.classification"
     _description = "Fiscal Classification"
 
-    name = fields.Char(required=True, translate=True)
+    name = fields.Char(required=True)
 
     description = fields.Text()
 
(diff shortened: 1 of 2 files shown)


legalsylvain (approved):
Hi @alexis-via. Thanks for the contribution ! 
1. commit about the view :+1: 
2. commit about the translation : 

- I personally don't use this module in a multicompany context. (I mean 1 classification per company no "global classification").
- I personally don't use translation. (all my companies are french).
- I so do what your propose in a "custom module" for the time being ;-) and I plan to port that module in V16 : https://github.com/grap/grap-odoo-custom/blob/12.0/grap_change_translation/models/account_product_fiscal_classification.py#L11

However, do you think it's valid for all the OCA users ? 

On the other hand, In odoo : 
- ``acount.tax`` is not translatable : https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_tax.py#L88
- ``account.tax.group`` is translatable : https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_tax.py#L26

In fact, I don't know if there is real configuration with global classification in a multi country context. @rvalyi, @PierrickBrun : a PoV on that topic ?

rvalyi:
Hello @legalsylvain, I confirm we designed this module for supporting global classification over several companies and we have a use case at Akretion with such a case on v14. It's an important feature I think.

legalsylvain:
I had understood that you had stopped using this module in Brazil? Did I miss something ?

PR #362, opened 2023-06-15
[IMP] account_avatax_oca: add search criteria to avoid singleton

JuanyDForgeflow:
Old behavior: by looking just with invoice doc number there's a chance to get a singleton on the search
Solution: search also by partner so we don't get a singleton issue

### DIFF
--- a/account_avatax_oca/models/account_move.py
+++ b/account_avatax_oca/models/account_move.py
@@ -176,7 +176,12 @@ class AccountMove(models.Model):
     # Same as v12
     def get_origin_tax_date(self):
         if self.invoice_doc_no:
-            orig_invoice = self.search([("name", "=", self.invoice_doc_no)])
+            orig_invoice = self.search(
+                [
+                    ("name", "=", self.invoice_doc_no),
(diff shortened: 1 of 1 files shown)


mariadforgeflow:
@JuanyDForgeflow  I think maybe you should search by name and journal, since the sequence is unique for each.

dreispt (approved):

account_avatax_oca/models/account_move.py
```diff
@@ -176,7 +176,12 @@ def onchange_reset_avatax_amount(self):
     # Same as v12
     def get_origin_tax_date(self):
         if self.invoice_doc_no:
-            orig_invoice = self.search([("name", "=", self.invoice_doc_no)])
+            orig_invoice = self.search(
+                [
+                    ("name", "=", self.invoice_doc_no),
+                    ("partner_id", "=", self.partner_id.id),
```
dreispt, line 182: Would this be better done by Commercial Partner ID?
(I'm aware Odoo OOB is not great at this, but still, let's not contribute to the problem...)

PR #366, opened 2023-07-10
[14.0][FIX]account_avatax_oca, fix invoice line amount with currency

atchuthan:
For multi currency scenario, the amount sent to Avatax is wrong.  If the amount is converted as per different currency, then the tax returned from Avatax should also be converted.  If we use amount_currency, the amount is converted properly to the currency's info and the currency is also sent to Avatax, so we don't need to convert the resulting tax amount.

cc: @SodexisTeam

### DIFF
--- a/account_avatax_oca/models/account_move.py
+++ b/account_avatax_oca/models/account_move.py
@@ -477,26 +477,7 @@ class AccountMoveLine(models.Model):
         adjusted to compute line amount instead of unit price.
         """
         self.ensure_one()
-        base_line = self
-        move = base_line.move_id
-        sign = -1 if move.is_inbound() else 1
-        quantity = qty or base_line.quantity
-        base_amount = base_line.price_unit * quantity
-        if base_line.currency_id:
-            price_unit_foreign_curr = (
-                sign * base_amount * (1 - (base_line.discount / 100.0))
-            )
-            price_unit_comp_curr = base_line.currency_id._convert(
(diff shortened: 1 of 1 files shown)


account_avatax_oca/models/account_move.py
```diff
@@ -477,26 +477,7 @@ def _get_avatax_amount(self, qty=None):
...
-                move.company_id.currency_id,
-                move.company_id,
-                move.date,
-            )
-        else:
-            price_unit_comp_curr = (
-                sign * base_amount * (1 - (base_line.discount / 100.0))
-            )
-        return -price_unit_comp_curr
+        return -self.amount_currency
```
dreispt, line 480: I'm unsure with this, as it is removing the discount handling logic.
There is the risk that discounts are not processed properly after this change.

atchuthan: @dreispt 
discount
avatax1

The amount_currency field has the discounted amount already.

dreispt: @atchuthan My concern is with the Avatax API call sending wrong amounts. Can you double check that?

atchuthan: @dreispt Yes, our customer is using this modified code for few months, and they haven't noticed any problem with the transactions

P.S: Sorry for the late reply.  Is it possible to reopen this PR?

PR #367, opened 2023-07-12
[15.0][MIG] l10n_eu_oss_oca: Set country_id for previous taxes

P4R:
Odoo adds country_id in account_tax in version 15.0 and if you come from previous version of Odoo country_id is set by migration of account module and set country of company. But for OSS taxes it's wrong. The correct country_id is the same as oss_country_id

### DIFF
--- a/l10n_eu_oss_oca/migrations/15.0.2.2.0/post-migration.py
+++ b/l10n_eu_oss_oca/migrations/15.0.2.2.0/post-migration.py
@@ -1,0 +1,16 @@
+# Copyright 2021 Sygel - Manuel Regidor
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade  # pylint: disable=W7936
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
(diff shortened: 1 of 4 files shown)


P4R:
Someone could review this? @HaraldPanten @pedrobaeza @RodrigoBM

RodrigoBM (approved):

dreispt:
If this is a [MIG] migration, why does it have only one commit?
I expected to see a commit history.
Or is it a forward port of another PR?

l10n_eu_oss_oca/migrations/15.0.2.2.0/post-migration.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2021 Sygel - Manuel Regidor
```
ValentinVinagre, line 1: I think the copyright is not correct.

P4R: Yes, this is wrong, I copied it by error. But now I don't have access to this branch because I leaved these organization.

ValentinVinagre: In that case you would have to cancel the PR and make the change in one that you have access to so that it is not accepted by mistake.

PR #370, opened 2023-08-02
[16.0][MIG] account_avatax_repair_oca: Migration to 16.0

AlexPForgeFlow:
Migration to 16.0


AlexPForgeFlow:
@legalsylvain Tests fail here https://github.com/OCA/account-fiscal-rule/actions/runs/5807941000/job/15743728976?pr=370#step:8:221 due to module [account_product_fiscal_classification](https://github.com/OCA/account-fiscal-rule/tree/16.0/account_product_fiscal_classification), and the module is completely unrelated. Not sure how to avoid this.

PR #275, opened 2022-03-21
[FIX] account_product_fiscal_classification: order by fiscal classification by name

legalsylvain:
trivial patch. order by name, and not by id.

CC : @rvalyi.

GRAP-403

PR #379, opened 2023-10-23
[16.0][IMP] account_product_fiscal_classification: add sequence

alexis-via:


account_product_fiscal_classification/models/account_product_fiscal_classification_template.py
```diff
@@ -8,6 +8,7 @@
 class AccountProductFiscalClassificationTemplate(models.Model):
     _name = "account.product.fiscal.classification.template"
     _description = "Fiscal Classification Template"
+    _order = "sequence, id"
```
legalsylvain:
```suggestion
    _order = "sequence, name"
```

if two items have the same sequence. (that occures, when updating this module), better to order by name.

account_product_fiscal_classification/models/account_product_fiscal_classification.py
```diff
@@ -9,6 +9,7 @@
 class AccountProductFiscalClassification(models.Model):
     _name = "account.product.fiscal.classification"
     _description = "Fiscal Classification"
+    _order = "sequence, id"
```
legalsylvain:
```suggestion
    _order = "sequence, name"
```
same here.

account_product_fiscal_classification/models/account_product_fiscal_classification.py
```diff
@@ -20,6 +21,8 @@ class AccountProductFiscalClassification(models.Model):
         " Classification without removing it.",
     )
 
+    sequence = fields.Integer(sequence=10)
```
legalsylvain:
```suggestion
    sequence = fields.Integer(default=10)
```

typo.

legalsylvain (changes requested):
minor remarks. Otherwise, LGTM.
thanks !

account_product_fiscal_classification/models/account_product_fiscal_classification_template.py
alexis-via: In the current implementation, there is no "_order", so it is order by id. So "sequence, id" keeps the current order on existing installations.

legalsylvain: indeed, but is it what we want ?

alexis-via: yes ! That way, we don't change the habits of the users until they decide the exact order they want.

legalsylvain: I mean, order by id is wrong, and has been fixed by https://github.com/OCA/account-fiscal-rule/pull/275 monthes ago. 
But this patch has not been back-ported.
I think it should. 
the order by sequence is OK, but please take the opportunity to backport this trivial patch !

thanks !

alexis-via:
On the form view of account.product.fiscal.classification, we have a "notebook" with a single "page"... why ? There are other modules which add tabs to this form view ?

legalsylvain:
> we have a with a single ... why ? 

?

alexis-via:
comment updated. Github dislikes when you write XML tags directly in the text of the comment!

alexis-via:
Ready to merge !

dreispt:
@legalsylvain Can you approve please, so we follow due process?

legalsylvain:
Hi ! 

> On the form view of account.product.fiscal.classification, we have a "notebook" with a single "page"... why ? There are other modules which add tabs to this form view ?

Generally, I add a notebook on form view. It allows to add extra tab in custom / OCA modules.

> @legalsylvain Can you approve please, so we follow due process?

No, my comment (https://github.com/OCA/account-fiscal-rule/pull/379#discussion_r1370332588) is not answered.

But no problem : I backported #275 here (#388), so this PR has just to be rebase, and could be merged.

alexis-via:
rebased

PR #390, opened 2023-11-09
[MOD]account_avatax_oca: Make a rounding tax from 4 to 2

sandipvyas-osi:
[MOD]account_avatax_oca: Make a rounding tax from 4 to 2


dreispt (changes requested):
@sandipvyas-osi The point for how it is working was to minimize the differences between the Avatax amount computation and the Odoo one.
Reducing the decimals will cause more differences and is not helpful.
Is there a particular reason to propose this change?

PR #399, opened 2023-12-12
[16.0][FIX] l10n_eu_oss_oca: Set tax country_id as company's country

aritzolea:
OSS taxes' `country_id` field was being filled with the same value as `oss_country_id` field.

This is causing some errors when managing invoices which are solved by setting taxes' company's country value to the tax `country_id` field.


l10n_eu_oss_oca/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 
 {
     "name": "L10n EU OSS OCA",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.0.1",
```
dreispt, line 6: No need to do this, as the ocabot will bump the version number.

dreispt (approved):

ghost (approved):
LGTM!

aritzolea:
@dreispt Looks like merge command didn't work. Could you try again?

dreispt:
@aritzolea Can you check https://github.com/OCA/account-fiscal-rule/pull/367?
It it the same issue?
Should this be backported, or should that one be merged?

HaraldPanten:
@aritzolea @pedrobaeza I'm not sure if these changes are creating tests failures in our 16.0 branch for Spanish location. look at these failures, for example --> https://github.com/OCA/l10n-spain/pull/3281

pedrobaeza:
Yes, this PR shouldn't be merged without a real explanation of what is fixing, and preferably a regression tests. The text "some errors" is not valid. You have to specify exactly the steps to reproduce the problem.

@aritzolea please fix the derived problem, or this PR will be reverted tomorrow until a better explanation (and solution) is provided.

SoniaViciana:
Hello @pedrobaeza 

If for any reason the taxes on an invoice are wrong and you want to correct them by changing the fiscal position, without this fix, the following error will appear:

https://www.awesomescreenshot.com/video/23476968?key=ec856d72b95896ef2ee940d3bf1d0895

ValentinVinagre:
> Hello @pedrobaeza
> 
> If for any reason the taxes on an invoice are wrong and you want to correct them by changing the fiscal position, without this fix, the following error will appear:
> 
> https://www.awesomescreenshot.com/video/23476968?key=ec856d72b95896ef2ee940d3bf1d0895

We understand the error, but the fix was not done correctly.

pedrobaeza:
After seeing the video, I have to say that the fix is correct indeed. Thanks for the extra info. Having this kind of info from the beginning helps to trace the source of the problem. 

In a nutshell, `country_id` is a core field for setting the country in which the tax is applicable. This has been introduced for multi-vat companies. It should be the company one, as you are applying such tax in your country, not on the destination country (that's not OSS!).

Being strict, the patch is not completely correct, as instead of assigning `country_id` from company, it should assign `account_fiscal_country_id`, which is the proper fiscal country. Most of the times, it will be the same, but not always, so we should put it correctly. Check that this is done in core module: https://github.com/odoo/odoo/blob/29aedbb3f7f7627723f0319d60730ef51521c90c/addons/l10n_eu_oss/models/res_company.py#L33C45-L33C70

Please do a PR changing such field in https://github.com/OCA/account-fiscal-rule/pull/399/files#diff-bc42cace184023a2d65a73e9c8af20970313cfbdf6380f58cb6d3450ba771451R119 and in the migration script.

Thus, I'm going to check what is happening in subsequent modules.

PR #401, opened 2023-12-14
[16.0][FIX] account_avatax_sale_oca: Fix record id

ArnauCForgeFlow:
Fixed record id, without this change the module wasn't displaying the inherited view


dreispt:
@ArnauCForgeFlow I don't understand this fix. If the original XML ID doesn't work, why would the changed XML ID work?

PR #402, opened 2023-12-22
[16.0][FIX] l10n_eu_oss_oca: Set tax country_id as company's fiscal country

aritzolea:


pedrobaeza:
The previous PR wasn't rollbacked, so you only need to change the field.

aritzolea:
But if someone updated to `16.0.1.0.1`, a new `16.0.1.0.2` version script is needed so that existing taxes are updated on the database right?

pedrobaeza:
Just modify previous migration script instead of creating a new one. The possibilities of being different are few, and also the possibilities of people updating in this lapse, so let's reduce the noise.

l10n_eu_oss_oca/README.rst
```diff
@@ -7,7 +7,7 @@ L10n EU OSS OCA
    !! This file is generated by oca-gen-addon-readme !!
    !! changes will be overwritten.                   !!
    !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
-   !! source digest: sha256:fde85cb88c420131b91b360fbb84f553d4977cbbfc80cc868c52291833e08e03
+   !! source digest: sha256:9f24da0b57196401086a0a54b7bb7e24024c260cc4c90d42308810512634a55c
```
pedrobaeza: Why this change and the one in the index.html?

aritzolea: Were residual changes from the first pre-commit when I added a new migration script

pedrobaeza (approved):
/ocabot merge patch

PR #404, opened 2023-12-26
[16.0][FIX] account_avatax_oca, skip exception raise on lines with not product in display_type

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/account_avatax_oca/hooks.py
+++ b/account_avatax_oca/hooks.py
@@ -1,6 +1,11 @@
 # Copyright (C) 2022 Open Source Integrators
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+from odoo import _, api
+from odoo.tools import frozendict
+
+from odoo.addons.account.models.account_move_line import AccountMoveLine
+
 
 def pre_init_hook(cr):
     # Preserve key data when moving from account_avatax to account_avatax_oca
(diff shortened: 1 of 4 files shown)


account_avatax_oca/models/account_tax.py
```diff
@@ -82,26 +82,39 @@ def compute_all(
...
+                        avatax_invoice.company_id.currency_id,
+                        avatax_invoice.company_id,
+                        avatax_invoice.date,
+                    )
+                    if (
+                        line.product_id == product
+                        and float_compare(line.quantity, quantity, digits) == 0
+                    ):
+                        avatax_amount = copysign(line.avatax_amt_line, base)
+                        break
```
dreispt, line 117: A lot of logic changes here that I don't understand , I'm not comfortable with this.

account_avatax_oca/hooks.py
```diff
@@ -13,3 +18,85 @@ def pre_init_hook(cr):
...
+                if tax["amount"]
+            }
+            if not line.tax_repartition_line_id:
+                line.compute_all_tax[frozendict({"id": line.id})] = {
+                    "tax_tag_ids": [(6, 0, compute_all_currency["base_tags"])],
+                }
+
+    if not hasattr(AccountMoveLine, "_compute_all_tax_origin"):
+        AccountMoveLine._compute_all_tax_origin = AccountMoveLine._compute_all_tax
+    AccountMoveLine._compute_all_tax = _compute_all_tax_new
```
dreispt, line 102: It looks like we are monkey patching the _compute_all_tax method.
Why would we need to do that?

dreispt (changes requested):

account_avatax_oca/hooks.py
ChrisOForgeFlow: It's needed because in this line https://github.com/odoo/odoo/blob/16.0/addons/account/models/account_move_line.py#L940, I'm adding a context with current aml, to know what account move line trigger tax compute, then in next process I can process with more precision

account_avatax_oca/models/account_tax.py
ChrisOForgeFlow: This change is only for adding a process of current aml, if that is not found, should be done standard behavior, you can notice change in showed only for tab adding in else condition

kobros-tech (changes requested):
I reviewed the module by running tests, you may need to add this code:
`
cls.journal = cls.env['account.journal'].create({
            'name': 'Test Sales Journal',
            'type': 'sale',
            'code': 'TSJ',
            'company_id': cls.company2.id,
        })
`
in tests/test_account_tax.py after cls.company2

with this tests will succeed :)

kobros-tech:
updated to #469

AlexPForgeFlow:
@kobros-tech Thanks for the improvement. Could you conserve original @ChrisOForgeFlow commit? In the OCA we typically conserve authorship of commits and history log. Then you can add a dedicated commit with your changes.

kobros-tech:
> @kobros-tech Thanks for the improvement. Could you conserve original @ChrisOForgeFlow commit? In the OCA we typically conserve authorship of commits and history log. Then you can add a dedicated commit with your changes.

@AlexPForgeFlow 
I can not add any commit on top of not merged one, unless I have access to the original branch.

We are using many of your open source, and I preserve your effort.

I only append my name if I make a lot of changes.

Even more the original PR is there, you can copy what I have fixed and get your PR merged, I don't have any problem with that as it is not the first time or the last time to help someone with their work.

I can archive this PR once the original one is merged.

AlexPForgeFlow:
@kobros-tech I really appreciate your work in this PR and others in which you have contributed.

There are two ways you can conserve commit history and add your changes: 
1. Clone original branch, in this case (https://github.com/ForgeFlow/account-fiscal-rule/tree/16.0-fix-compute_all_exception), add your own commit or your changes and open a new PR on the OCA.
2. Open a PR against original branch, in this case at https://github.com/ForgeFlow/account-fiscal-rule

I hope you understand. Thanks.

PR #420, opened 2024-03-21
[16.0][IMP] l10n_eu_oss_oca: Support for regional tax rates, Portugal regions

hildickethan:
Add support for regional OSS taxes, in this case the Azores and Madeira regions in Portugal.
Mentioned in the original PR https://github.com/OCA/account-fiscal-rule/pull/207#issuecomment-860566230 but was left pending

I checked other OCA modules that depend on this one and it doesn't seem like having multiple taxes/fiscal positions per country will break anything. I did a small test with a 303 report (`l10n_es_aeat_mod303_oss`) and it added the new taxes with no issue

ping @dreispt since you originally wanted this in 13.0


dreispt:
@hildickethan-S73 Thank you for working on this - that discussion was a long time ago 😄 
I don't see the Spanish Islands here - I would like conformation this approach is also compatible it the way the Spanish localization works.

hildickethan:
> @hildickethan-S73 Thank you for working on this - that discussion was a long time ago 😄 I don't see the Spanish Islands here - I would like conformation this approach is also compatible it the way the Spanish localization works.

I didn't add them because right now I only need the PT ones. I tried it with a one of the Spanish accounting reports and it was OK, but it would be good to get input from @pedrobaeza too

pedrobaeza:
Canary Islands don't apply OSS.

Jortolsa-S73 (approved):

fuentes73 (approved):

hildickethan:
This PR has been working well for us @dreispt, it should be OK to merge

l10n_eu_oss_oca/data/oss.tax.rate.csv
```diff
@@ -1,28 +1,30 @@
...
-oss_eu_rate_mt,base.mt,18,7,5,0
-oss_eu_rate_nl,base.nl,21,9,0,0
-oss_eu_rate_pl,base.pl,23,8,5,0
-oss_eu_rate_pt,base.pt,23,13,6,0
-oss_eu_rate_ro,base.ro,19,9,5,0
-oss_eu_rate_sk,base.sk,20,10,0,0
-oss_eu_rate_si,base.si,22,9.5,5,0
-oss_eu_rate_es,base.es,21,10,4,0
-oss_eu_rate_se,base.se,25,12,6,0
+id,oss_country_id:id,oss_state_ids:id,general_rate,reduced_rate,superreduced_rate,second_superreduced_rate
```
pedrobaeza, line 1: Can you avoid this big diff?

pedrobaeza: OK, nothing... there's an extra column.

PR #424, opened 2024-04-17
[FIX] Allowing Internal Users to Validate the Contact Address

tirth-osi:
Issue:
Go to Contacts>Avatax>Click on Validate
For a user with no Accounting Accountant Access we get Access Error.

Screenshot from 2024-04-17 14-42-04

This PR will ensure that any internal user is able to validate the address.


tirth-osi:
@dreispt Please review.

dreispt (approved):

PR #439, opened 2024-06-24
[14.0][IMP] add tax_scope to account.product.fiscal.classification

TB-Ph35:
add tax-scope to tax classification. 
Tax in tax classification are filter with the values of the tax_scope.
On product, the tax_classification are filter with the type of product(consu or service)


account_product_fiscal_classification/models/product_template.py
```diff
@@ -15,16 +15,32 @@ class ProductTemplate(models.Model):
...
 
+    # Depends Section
+    @api.depends("type")
+    def _compute_type_identification(self):
+        for product in self:
+            if product.type == "product":
+                product.tax_scope_identification = "consu"
+            else:
+                product.tax_scope_identification = product.type
+
```
legalsylvain, line 43: why add a new field on product.product. You could just base your domain on the product.type field ? 

thanks !

TB-Ph35: It was the why i found to have the product.type "product" and "consu" on one label.
it may be to restrictive. but i don't found a other idea to get this.

Sorry, for the time. I misclick my answer

PR #442, opened 2024-07-15
[16.0][FIX] account_avatax_sale_oca: Avoid reset of order line price after saving

JordiBForgeFlow:
We found a bug where the price when the decimal accuracy was 5, the price would be automatically reset to the pricelist price after saving the order. The reason was a wrong implementation of onchange_reset_avatax_amount. In this test we demonstrate that the error does not happen anymore.

The PR contains 2 commits:
* Commit 1 contains a test that reproduces the error
* Commit 2 contains the fix, and the test implemented in previous commit now passes

### DIFF
--- a/account_avatax_sale_oca/models/sale_order.py
+++ b/account_avatax_sale_oca/models/sale_order.py
@@ -89,7 +89,6 @@ class SaleOrder(models.Model):
         """
         for order in self:
             order.tax_amount = 0
-            order.order_line.tax_amt = 0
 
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
Failed test before fix https://github.com/OCA/account-fiscal-rule/actions/runs/9935890748/job/27442968925?pr=442#step:8:205

JordiBForgeFlow:
See how after the second commit the tests now pass.

dreispt (approved):

PR #444, opened 2024-08-13
[17.0][FIX] account_avatax_oca: Install on multi-country

RLeeGML:
On install of account_avatax_oca in an environment where the main company is not in the United States, install fails with error "The tax group must have the same country_id as the tax using it.". The default tax is created in the main company, and the country_id is defined in the XML data file, but the tax group computes its country_id from the main company's country, and the SQL constraint error appears. This is true even if the active company is a USA company, and the main company is not active.

### DIFF
--- a/account_avatax_oca/data/avalara_salestax_data.xml
+++ b/account_avatax_oca/data/avalara_salestax_data.xml
@@ -7,6 +7,7 @@
     </record>
     <record id="avatax_tax_group" model="account.tax.group">
         <field name="name">AvaTax</field>
+        <field name="country_id" ref="base.us" />
     </record>
     <!-- Used as template for automatic Tax records created -->
     <record id="avatax" model="account.tax">


dreispt:
I'm afraid this fix is just a workaround to allow you installing the module.
The country is needs to be set as US.
It is just that you don't want the Fiscal Position to be created for a non-US base company.

Perhaps the proper solution is to have this config data added though the CoA Template...

RLeeGML:
I agree that this is a workaround. The ideal situation would leave you with the module data created on the US company. However, I don't see a way that would cover every scenario (e.g. US company hasn't been created yet, main company is non-US but does sales in US), so I think the best solution may be to allow install, then instruct the user to use the module data as a template and correctly create the data in whatever company they need to. Unfortunately, that's not simple either, as tax groups cannot be created/imported through the UI.

PR #449, opened 2024-10-10
[FIX] Avatax: Use Odoo Tax option

AlexPForgeFlow:
Forward Port from commit: https://github.com/OCA/account-fiscal-rule/commit/4c56a80fdc65d96d101a2d60a91f113104eee798


AlexPForgeFlow:
Hi @atchuthan I cherry-picked your fix, but I have one question, I hope you can help me.
In your proposal, you use the field avatax_tax_type, to manage visibility for option "Use Odoo Tax". The problem I have is that for those invoices that taxes have already been computes if I want to create a credit note, option "Use Odoo Tax" would not be visible, and I don't want to recompute taxes for those invoices because taxes may have changed. My question is, there is some default code I can set for avatax_tax_type? Wouldn't be the same if instead of using avatax_tax_type for option visibility we use field is_avatax or check avatax_amt_line.

Thanks!

atchuthan:
> Hi @atchuthan I cherry-picked your fix, but I have one question, I hope you can help me. In your proposal, you use the field avatax_tax_type, to manage visibility for option "Use Odoo Tax". The problem I have is that for those invoices that taxes have already been computes if I want to create a credit note, option "Use Odoo Tax" would not be visible, and I don't want to recompute taxes for those invoices because taxes may have changed. My question is, there is some default code I can set for avatax_tax_type? Wouldn't be the same if instead of using avatax_tax_type for option visibility we use field is_avatax or check avatax_amt_line.
> 
> Thanks!

For old records of our customer, we didn't have the tax type, so we had to make code change to check if avatax_tax_type is set in order to work with "Use Odoo Tax".  I think we can change as per your suggestion and generalize them.

Thanks

PR #452, opened 2024-11-04
[FIX] account_fiscal_position_type: prevent error in demo / test mode, if l10n_generic_coa is not installed

legalsylvain:
**rational**

demo and test are based on l10n_generic_coa data. However, if other localization are installed, the installation of this module will fail with demo data.

**Action**
create a dedicated demo chart of account.

similar to v16 implementation.

https://github.com/OCA/account-fiscal-rule/blob/16.0/account_fiscal_position_type/demo/account_chart_template.xml

### DIFF
--- a/account_fiscal_position_type/__manifest__.py
+++ b/account_fiscal_position_type/__manifest__.py
@@ -21,6 +21,7 @@
     ],
     'demo': [
         'demo/res_groups.xml',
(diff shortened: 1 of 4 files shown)


rvalyi (approved):

legalsylvain:
@rvalyi : could you merge ?

dreispt (approved):

PR #457, opened 2024-12-03
[16.0][FIX] account_fiscal_position_type: use company from move

hbrunn:
otherwise you have to switch company for editing fiscal positions of moves from another company than the active one


dreispt (approved):
Looks good. I wonder if there are other places in the code where this fix might be needed.

dreispt:
For a more complete fix see #497

rvalyi (dismissed):

dreispt:
@rvalyi I'm actually wrong, I mixed up PRs, sorry!

PR #458, opened 2024-12-04
[17.0][MIG] account_avatax_oca_log: Migration to 17.0

kobros-tech:
@dnplkndll


wlin-kencove:
Runboat does not seem to be working so I am testing locally. 

I believe I had setup the configuration correclty:

image

wlin-kencove:
image


I am able to calculate sales tax and  get a rate but I do not see the avatax log tab nor do I see anything in the avatax logs:

image

wlin-kencove:
@kobros-tech  am I missing any configs or perhaps a bug?

kobros-tech:
> @kobros-tech am I missing any configs or perhaps a bug?

run bot was sick today, but I will review the migration as we focused on 18.0 and did many updates there .

kobros-tech:
> @kobros-tech am I missing any configs or perhaps a bug?

I reviewed the module and I am certain if you install account_avataxt_website_sale along with setting the log field true, the log will be created once you save the SO in the log tap there.

Screenshot from 2025-04-08 09-16-21

Screenshot from 2025-04-08 09-33-08

dnplkndll:
@kobros-tech should demo mode on a module enable the only feature of the module without manual config? or is that not the convention?

dnplkndll:
> @kobros-tech should demo mode on a module enable the only feature of the module without manual config? or is that not the convention?

along those lines. is there a convention on the documentation/usage to always call out the config to enable? I think that would be a useful thing to add to all the modules / migrations we complete.

kobros-tech:
if specified yes, but I don't see this feature here.

I think it is limited with the front end functions.

PR #462, opened 2024-12-06
[FIX] account_avatax_oca: only override exemption if provided in Odoo

dreispt:
If no exemption info is provided in Odoo, the API call sends None.
This causes and override on exemption information that might exist in
the Avatax side.


kobros-tech (approved):
we can add a test case for it in the next commit

kobros-tech:
@dreispt 

If you don't mind can I open this PR again so that it gets merged, I know you are too much busy and I like to help?

PR #467, opened 2025-01-07
[18.0][MIG] : account_avatax_website_sale Migration to 18.0

kobros-tech:
@dreispt 

This PR is dependent on #468 PR and before getting merged we need a clue to remove test-requirements dependency on the 'account_avatax_sale_oca' PR for migration to 18.0.

In fact I recommend having #468 get merged first, then we remove test dependency, and finally this PR gets merged.

To set avatax taxes in sale and website sale too:

https://github.com/OCA/account-fiscal-rule/pull/467#issuecomment-2637169842


kobros-tech:
@dnplkndll

wlin-kencove:
Configured and tested. 

I was able to get taxes on orders entered in the backend but was not able to get tax for the same product on the website on checkout.

kobros-tech:
> Configured and tested.
> 
> I was able to get taxes on orders entered in the backend but was not able to get tax for the same product on the website on checkout.

let me see

wlin-kencove:
image


Above you can see the tax calculates on the backend with order entry. 


When I  go to the website and add the item to cart to checkout, tax is not calculated.  I have added the shipping address at this pont as well. 

image

wlin-kencove:
Found the issue, required the calculate customer code option turned on for web customers (since they wouldn't have that set yet).

image

Might be worth putting a pending status on this first checkout page to help customers avoid confusion.  
Maybe need a similar 'Pending' status here instead of 0.00. We do need delivery method first before we can determine tax (origin vs destination taxablity)
image


Tax only becomes visible on the payment entry screen:
image

wlin-kencove (approved):
Figured out a configuration required for customer for web customers to auto generate customer code and auto validate address. 

Seems good otherwise. 

The only suggestion would be to perhaps display a 'Pending' state for the tax field until taxes can be calculated (after address entry and shipping terms selection).

kobros-tech:
I agree with you
so for other reviewers I recommend having these settings before committing their review in website sales.

1. Tax Calculation
Screenshot from 2025-02-05 18-06-49

2. Address Validation
Screenshot from 2025-02-05 18-07-04

3. Product Tax Codes
Screenshot from 2025-02-05 18-07-51

4. Product Avatax Code 
Screenshot from 2025-02-05 18-08-14

ChuwenSun (approved):
Tax calculation is responding correctly after the address is entered.

dnplkndll:
@dreispt please review and merge if possible. thanks

dreispt (approved):

PR #472, opened 2025-01-14
[17.0][MIG] account_avatax_exemption

dnplkndll:


kobros-tech:
@dnplkndll 
because this module has dependency we see error in dependency, and I am migrating the dependency module.

I guess if there are tests, they should be applied in the account_avatax_exemption_base module.

migration PR for its base is: #477

PR #473, opened 2025-01-15
[18.0][MIG] account_avatax_exemption_base: Migration to 18.0

dnplkndll:
@kobros-tech


wlin-kencove (changes requested):
image


image

Most of the fields look good. 

One thing I noticed is that the res.partner.exemption form view is overlapping with another form.

image

See image above. I am not able to save an exemption since I am not able to edit the required fields behind

kobros-tech:
@wlin-kencove 

Thanks very much for tracing this error, in fact technically it is not an error, as in this scenario you must enter a valid model name instead of "test" like "res.partner" and this will work fine.

kobros-tech:
To get rid of similar error, we can add a commit to handle selecting model name based on some domain or rule to avoid misspelling the model name.

here is my review result:

Screenshot from 2025-02-05 22-33-48

Screenshot from 2025-02-05 22-38-00

Screenshot from 2025-02-05 22-38-51

Screenshot from 2025-02-05 22-39-04

Screenshot from 2025-02-05 22-39-36

wlin-kencove:
@kobros-tech 

Here are the steps to get the error I was seeing:

 I search for avatax exemption direclty:
image

From there, I select to add new: 
image

And this view appears:. 
image


I can only enter into those lines but cannot edit the fields behind it. 
image


Please let me know if you can reproduce.

wlin-kencove:
@kobros-tech  here is the view when I enter via add exemption from the customer record, which looks correct:
image

Looks like maybe some spacing/sizing issues? 
image

kobros-tech:
@dnplkndll 
@wlin-kencove 

This is how the form looks like:

Screenshot from 2025-02-06 01-46-20

Screenshot from 2025-02-06 01-47-04

other features in this form are introduced in "account_avatax_exemption" dependent module.

wlin-kencove:
image


View looks good now.

wlin-kencove (approved):
View update looks good now without the components overlapping each other.

account_avatax_exemption_base/views/avalara_exemption_view.xml
```diff
@@ -214,11 +214,6 @@
                         </page>
                     </notebook>
                 </sheet>
-                <div class="oe_chatter">
-                    <field name="message_follower_ids" widget="mail_followers" />
-                    <field name="activity_ids" widget="mail_activity" />
-                    <field name="message_ids" widget="mail_thread" />
-                </div>
```
xaviedoanhduy: hi @kobros-tech  have you tried applying this change? see: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

> The `<div class="oe_chatter">` with all its inner stuff now converts into just <chatter />. See https://github.com/odoo/odoo/pull/156463 for more details and sample replacements.

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+
+class ResPartnerExemptionLine(models.Model):
+    _name = "res.partner.exemption.line"
+    _description = "Avatax Exemption Line"
+
+    name = fields.Char(index=True, default=lambda self: _("New"))
+    exemption_id = fields.Many2one("res.partner.exemption", required=True)
+    partner_id = fields.Many2one(related="exemption_id.partner_id", store=True)
+    state_id = fields.Many2one("res.country.state")
+    avatax_id = fields.Char("Avatax Certificate ID", copy=False, readonly=True)
```
xaviedoanhduy, line 25: > If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead

see: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+            if record.exemption_number:
+                name = f"{record.exemption_number} - {record.partner_id.display_name}"
+            else:
+                name = record.partner_id.display_name
+            if record.exemption_type:
+                name = f"{record.exemption_type.name} - {name}"
+            res.append((record.id, name))
+        return res
+
+    # TODO: Need to check, avalara.salestax not found and also field.
```
xaviedoanhduy: fyi, `exemption_export` is declared in the `account_avatax_exemption` module, see: https://github.com/OCA/account-fiscal-rule/blob/8f0093f77fa4cae161f75d30e4ae71d8013a64f4/account_avatax_exemption/models/avalara_salestax.py#L18

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+    state_ids = fields.Many2many("res.country.state", string="States")
+    exemption_validity_duration = fields.Integer(
+        help="Validity duration in days", default=30
+    )
+
+    @api.onchange("group_of_state")
+    def onchange_group_of_state(self):
+        if self.group_of_state.state_ids and not self.state_ids:
+            self.state_ids = [(6, 0, self.group_of_state.state_ids.ids)]
```
xaviedoanhduy, line 69:
```suggestion
            self.state_ids = [Command.set(self.group_of_state.state_ids.ids)]
```
using `Command` is IMHO better, see: https://www.odoo.com/documentation/17.0/developer/reference/backend/orm.html#odoo.fields.Command

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+    def _get_document_folder(self):
+        return self.env.company.documents_exemption_folder
+
+    def _check_create_documents(self):
+        return (
+            self.env.company.documents_exemption_settings
+            and super()._check_create_documents()
+        )
+
+    def name_get(self):
```
xaviedoanhduy: > If you are overriding name_get method, you should now override _compute_display_name instead, adding the possible new fields in the depends of the method. See https://github.com/odoo/odoo/pull/122085 for more details.

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+            self.group_of_state = self.exemption_type.group_of_state.id
+        if self.exemption_type or self.group_of_state:
+            state_ids = []
+            if self.exemption_type.group_of_state.state_ids:
+                state_ids += self.exemption_type.group_of_state.state_ids.ids
+            if self.exemption_type.state_ids:
+                state_ids += self.exemption_type.state_ids.ids
+            if self.group_of_state.state_ids:
+                state_ids += self.group_of_state.state_ids.ids
+            self.state_ids = [(6, 0, list(set(state_ids)))]
```
xaviedoanhduy: better with `Command`

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+            self.expiry_date = self.effective_date + timedelta(
+                days=self.exemption_type.exemption_validity_duration
+            )
+
+    @api.onchange("state_ids")
+    def onchange_state_ids(self):
+        if not any(self.exemption_line_ids.mapped("avatax_id")) and not any(
+            self.exemption_line_ids.mapped("add_exemption_number")
+        ):
+            self.exemption_line_ids = [(6, 0, [])]
```
xaviedoanhduy, line 156:
```suggestion
            self.exemption_line_ids = [Command.clear()]
```

account_avatax_exemption_base/models/exemption.py
```diff
@@ -0,0 +1,179 @@
...
+        for state_id in self.state_ids.ids:
+            if state_id not in self.exemption_line_ids.mapped("state_id").ids:
```
xaviedoanhduy, line 158:
```suggestion
        existing_state_ids = self.exemption_line_ids.mapped("state_id").ids
        for state_id in self.state_ids.ids:
            if state_id not in existing_state_ids:
```
avoid accesses in a loop

account_avatax_exemption_base/views/avalara_exemption_view.xml
```diff
@@ -0,0 +1,274 @@
...
+        <field name="model">res.partner.group.state</field>
+        <field name="type">list</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="name" />
+                <field name="country_id" />
+                <field
+                    name="state_ids"
+                    widget="many2many_tags"
+                    domain="[('country_id','=',country_id),('avatax_nexus', '=', True)]"
```
xaviedoanhduy: This attribute seems a bit redundant on a `list` view without the `editable` attribute.

account_avatax_exemption_base/views/avalara_exemption_view.xml
```diff
@@ -0,0 +1,274 @@
...
+        <field name="name">res.partner.exemption.type.tree.view</field>
+        <field name="model">res.partner.exemption.type</field>
+        <field name="type">list</field>
+        <field name="arch" type="xml">
+            <list>
+                <field name="name" />
+                <field
+                    name="state_ids"
+                    widget="many2many_tags"
+                    domain="[('avatax_nexus', '=', True)]"
```
xaviedoanhduy: the same above

account_avatax_exemption_base/views/avalara_exemption_view.xml
```diff
@@ -0,0 +1,274 @@
...
+                        name="state"
+                        widget="statusbar"
+                        statusbar_visible="draft,progress,done"
+                    />
+                </header>
+                <sheet>
+                    <group>
+                        <field
+                            name="partner_id"
+                            options="{'always_reload': True}"
```
xaviedoanhduy:
```suggestion
```
see: https://github.com/odoo/odoo/pull/130169/commits/a6673f67654a557494b57e6b7112d25818dde6e3

account_avatax_exemption_base/views/avalara_exemption_view.xml
```diff
@@ -0,0 +1,274 @@
...
+                            invisible="context.get('hide_partner', False)"
+                            readonly="state != 'draft'"
+                        />
+                        <field name="exemption_type" readonly="state != 'draft'" />
+                        <field
+                            name="business_type"
+                            options="{'no_create': True, 'always_reload': True}"
```
xaviedoanhduy:
```suggestion
                            options="{'no_create': True}"
```

account_avatax_exemption_base/views/avalara_exemption_view.xml
```diff
@@ -0,0 +1,274 @@
...
+                        />
+                        <field
+                            name="effective_date"
+                            required="1"
+                            readonly="state != 'draft'"
+                        />
+                        <field
+                            name="expiry_date"
+                            required="1"
+                            readonly="expiry_date != False and state != 'draft' or expiry_date != False"
```
xaviedoanhduy:
```suggestion
                            readonly="expiry_date or state != 'draft'"
```

xaviedoanhduy (changes requested):

xaviedoanhduy (commented):
Thank you all for the above work but still need some minor changes above

Strangely `avatax_id` is a `readonly` field but is `required` and not automatically compute. How can I create data on this model if the above attributes are mutually exclusive.

image

xaviedoanhduy:
> @kobros-tech here is the view when I enter via add exemption from the customer record, which looks correct: image
> 
> Looks like maybe some spacing/sizing issues? image

I guess the problem is that `chapter` is not declared properly causing the inherited fields of this model to be displayed in normal form instead of components.

wlin-kencove:
> Thank you all for the above work but still need some minor changes above
> 
> Strangely `avatax_id` is a `readonly` field but is `required` and not automatically compute. How can I create data on this model if the above attributes are mutually exclusive.
> 
> image


You actually should not create exemption.business.type manually but pull them from an action that exports the records from a avalara enviorment. Below is the function from a previous version (16.0). 
image

This is not to be confused with avatax_id ids from other models as they store difference external references(ie: exemption, res.partner, etc).

xaviedoanhduy:
> You actually should not create exemption.business.type manually but pull them from an action that exports the records from a avalara enviorment. Below is the function from a previous version (16.0). 
> This is not to be confused with avatax_id ids from other models as they store difference external references(ie: exemption, res.partner, etc).

okay thanks. I tested it when applying it to dependent modules like `account_avatax_sale_oca`, `account_avatax_exemption` and `account_avatax_oca` and it seems to work pretty similar to the previous version (`16.0`). but the tweaks in my suggestion above should still be applied to make it more complete. 

Functionality LGTM
image

account_avatax_exemption_base/views/avalara_exemption_view.xml
dnplkndll: You are saying if not editable, then there are only set values no need to filter available correct?

xaviedoanhduy: yes indeed, my opinion is that it should be removed because we are actually only operating in read mode instead of writing.
Just a personal opinion, you may agree or disagree

xaviedoanhduy:
this comment is still pending https://github.com/OCA/account-fiscal-rule/pull/473#discussion_r1957631435

account_avatax_exemption_base/views/avalara_exemption_view.xml
dnplkndll: I think the idea here is the chatter was not intended

xaviedoanhduy: I think you should drop this commit `[IMP] account_avatax_exemption_base: remove chatter from wizard view`.

replace the old syntax `<div class="oe_chatter">...</div>` to `<chatter />` include in Migration commit `[MIG] account_avatax_exemption_base: Migration to 18.0`. Because it is part of the migration process.
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-18.0

xaviedoanhduy (approved):
thanks for your patience. now it looks pretty good to me

xaviedoanhduy:
guys, should we take this opportunity to add a display name for this `res.partner.exemption`:

image

kobros-tech:
> guys, should we take this opportunity to add a display name for this `res.partner.exemption`:
> 
> image

It should be reviewed in the dependent module in this PR https://github.com/OCA/account-fiscal-rule/pull/475

And the name of the partner is displayed, try to review again  and if you encounter the issue ping us.

Screenshot from 2025-03-13 17-22-04

Screenshot from 2025-03-13 17-25-56

dnplkndll:
@dreispt this is ready for review / merge. Thanks for checking!

dreispt:
Please remove the test-requirements.txt before we merge.

kobros-tech:
> Please remove the test-requirements.txt before we merge.

I will now

kobros-tech:
@dreispt 

this base module doesn't have test requirements 

I am removing test requirements from its dependent one, I will drop .gitattributes file

kobros-tech:
@dreispt 
I just squashed a commit so we have the [MIG] commit only

PR #475, opened 2025-01-15
[18.0][MIG] account_avatax_exemption: Migration to 18.0 

dnplkndll:
- [ ] https://github.com/OCA/account-fiscal-rule/pull/473


dnplkndll:
@atchuthan can you look over? I tried to fix to get any commits from you guys in from 17. testing my git skills though and overwrote it once!

kobros-tech:
@dnplkndll 
@wlin-kencove

Based on #473 the form view should look like:

Screenshot from 2025-02-06 02-40-51

Screenshot from 2025-02-06 03-32-03

Screenshot from 2025-02-06 02-41-03

dnplkndll:
@benstannard since we introduce the exemption rules here. should we detail on the description the rules around SST compliance? don't want to have anyone get into trouble by misapplying those if participating in the program. 
I wonder if useful to ad a general SST compliance check would be useful. since you would need the members enabled? that would be more applicable on the base tax system module for states and I guess the new nexus check is not in odoo code now?

dnplkndll:
@kobros-tech @xaviedoanhduy @wlin-kencove 
we ready for reviews here?

kobros-tech:
> @kobros-tech @xaviedoanhduy @wlin-kencove we ready for reviews here?

I guess we shall make test cases for that module concerning exemption to simulate avalara APIs calls

account_avatax_exemption/models/exemption.py
```diff
@@ -0,0 +1,317 @@
...
+    def export_exemption(self):
+        self.ensure_one()
+        avalara_salestax = (
+            self.env["avalara.salestax"]
+            .sudo()
+            .search([("exemption_export", "=", True)], limit=1)
+        )
+        if not avalara_salestax:
+            raise UserError(
+                _("Avatax Exemption export is disabled in Avatax configuration")
```
xaviedoanhduy:
```suggestion
                self.env._("Avatax Exemption export is disabled in Avatax configuration")
```

account_avatax_exemption/models/exemption.py
```diff
@@ -0,0 +1,317 @@
...
+    taxable = fields.Boolean()
+
+    @api.constrains("avatax_rate")
+    def _check_avatax_rate(self):
+        """
+        Prevent the Avatax rate with range 0 to 100
+        """
+        for record in self:
+            if record.avatax_rate < 0 or record.avatax_rate > 100:
+                raise ValidationError(_("Avatax rate range is from 0 to 100"))
```
xaviedoanhduy:
```suggestion
                raise ValidationError(self.evn._("Avatax rate range is from 0 to 100"))
```

> You can replace the call to translate a string (`_`) with `self.env._` for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

Similar to other cases in this module

account_avatax_exemption/models/exemption.py
```diff
@@ -0,0 +1,317 @@
...
+        return True
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.model
+    def _search(self, domain, offset=0, limit=None, order=None):
+        context = dict(self._context)
+        if context.get("partner_exemption", False):
```
xaviedoanhduy:
```suggestion
        if self._context.get("partner_exemption", False):
```
it is always `dict`

account_avatax_exemption/models/exemption.py
```diff
@@ -0,0 +1,317 @@
...
+        return super()._search(domain, offset, limit, order)
+
+
+class ResPartnerExemption(models.Model):
+    _inherit = "res.partner.exemption"
+
+    exemption_code_id = fields.Many2one(
+        related="business_type.exemption_code_id",
+        string="Entity Use Code",
+        readonly=True,
```
xaviedoanhduy:
```suggestion
```
A `related` field is always read-only

Similar to other cases in this module

account_avatax_exemption/models/product.py
```diff
@@ -0,0 +1,108 @@
...
+        return res
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    avatax_item_id = fields.Char(
+        "Avatax TaxItem",
+        copy=False,
+        readonly=True,
```
xaviedoanhduy: > If a field is marked as readonly=True on the model, it won't be possible to import it through the Odoo import tool, so avoid it as possible and define the readonly attribute on the views instead.

see: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

account_avatax_exemption/views/avalara_exemption_view.xml
```diff
@@ -0,0 +1,200 @@
...
+        <field name="type">form</field>
+        <field
+            name="inherit_id"
+            ref="account_avatax_exemption_base.exemption_form_view"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='state']" position="before">
+                <button
+                    string="Export Exemptions"
+                    invisibl="state != 'draft'"
```
xaviedoanhduy:
```suggestion
                    invisible="state != 'draft'"
```
typo. replace all `invisibl` to `invisible` for all

account_avatax_exemption/views/exemption_template_views.xml
```diff
@@ -0,0 +1,84 @@
...
+                                </t>
+                                <t t-else="">
+                                    <t t-esc="'DISABLED'" />
+                                </t>
+                            </td>
+                        </tr>
+                    </tbody>
+                </table>
+               </t> -->
+                <t t-esc="message" />
```
xaviedoanhduy:
```suggestion
                <t t-out="message" />
```

Deprecated output directives, see: https://www.odoo.com/documentation/18.0/developer/reference/frontend/qweb.html#deprecated-output-directives

account_avatax_exemption/views/partner_view.xml
```diff
@@ -0,0 +1,36 @@
...
+        <field
+            name="inherit_id"
+            ref="account_avatax_oca.view_partner_details_form_inherit1"
+        />
+        <field name="arch" type="xml">
+            <xpath expr="//field[@name='customer_code']/.." position="after">
+                <group
+                    col="2"
+                    colspan="4"
+                    invisible="not is_company and parent_id != False"
```
xaviedoanhduy:
```suggestion
                    invisible="not is_company and parent_id"
```

account_avatax_exemption/views/partner_view.xml
```diff
@@ -0,0 +1,36 @@
...
+                >
+                    <field name="avatax_id" />
+                    <field name="id" invisible="1" />
+                    <field name="is_company" invisible="1" />
+                    <label for="exemption_ids" />
+                </group>
+                <field name="exemption_ids" context="{'hide_partner':id}" nolabel="1" />
+                <div invisibl="is_company or not parent_id">
```
xaviedoanhduy:
```suggestion
                <div invisible="is_company or not parent_id">
```

account_avatax_exemption/views/website_exemption_form.xml
```diff
@@ -0,0 +1,168 @@
...
+            </div>
+        </div>
+    </template>
+    <template id="form_field" name="Portal Exemption Form Field">
+        <div class="form-group">
+            <div class="row">
+                <div class="col-3">
+                    <span class="font-weight-bold" t-esc="'Business Type:'" />
+                </div>
+                <div class="col-9 text-left">
```
xaviedoanhduy:
```suggestion
                <div class="col-9 text-start">
```
replace all `text-left` to `text-start` and `text-right` to `text-end`

account_avatax_exemption/views/website_exemption_form.xml
```diff
@@ -0,0 +1,168 @@
...
+                    <span t-field="exemption.business_type" />
+                </div>
+            </div>
+            <div class="row mt16">
+                <div class="col-3">
+                    <span class="font-weight-bold" t-esc="'Entity Use Code:'" />
```
xaviedoanhduy:
```suggestion
                    <span class="fw-bold" t-out="'Entity Use Code:'" />
```
replace all `font-weight-bold` to `fw-bold`

account_avatax_exemption/views/website_exemption_tree.xml
```diff
@@ -0,0 +1,80 @@
...
+                                t-attf-href="/my/exemptions/#{slug(exemption)}?#{keep_query()}"
+                            >
+                                <span t-esc="'View'" />
+                            </a>
+                        </td>
+                    </tr>
+                </t>
+            </tbody>
+        </table>
+        <div t-if="pager" class="o_portal_pager text-center">
```
xaviedoanhduy:
```suggestion
        <div t-if="pager" class="o_portal_pager d-flex justify-content-center">
```
ref: https://github.com/odoo/odoo/blob/a0205ffa3d4a42321a41503b8056a9f1c89328bb/addons/website/views/website_templates.xml#L2697

xaviedoanhduy (changes requested):

account_avatax_exemption/controller/main.py
```diff
@@ -0,0 +1,166 @@
...
+                ("create_date", "<", date_end),
+            ]
+
+        # Make pager
+        pager = request.website.pager(
+            url=base_url,
+            url_args={"date_begin": date_begin, "date_end": date_end, "sortby": sortby},
+            total=Exemption.search_count(domain),
+            page=page,
+            step=self._items_per_page,
```
xaviedoanhduy, line 92:
```suggestion
            step=self._items_per_page,
```

After this commit https://github.com/odoo/odoo/pull/74996/commits/8a5e336ad31  the number of items per page has increased from `20` to `80`. We should consider adjusting this number to reduce the pressure on the database and improve the UI. Because if 80 items on 1 page is probably too much information.

xaviedoanhduy (changes requested):

account_avatax_exemption/views/website_exemption_tree.xml
xaviedoanhduy: image

xaviedoanhduy:
could you check this error: `Uncaught Promise > Invalid props for component 'WarningDialog': 'message' is not a string`

[Screencast from 06-03-2025 16:21:25.webm](https://github.com/user-attachments/assets/0ddeb084-1b2c-443c-bca4-e6316ed60106)

account_avatax_exemption/controller/main.py
dnplkndll: what is the convention to update? set self._items_per_page = 20 ?

account_avatax_exemption/views/partner_view.xml
wlin-kencove: just wanted to bring up that those two are not equal statements
 
 
for example in the original:
 
not is_company(False) and parent_id(None) != False  resolves to True


with change:
 
not is_company(False) and parent_id(None)  resolves to None -- i assume that this just results to 0 or false when set on "invisible"

I think the change would be more correct than the original assuming that we want to show the label if record is not company and record has not parent.

kobros-tech:
> could you check this error: `Uncaught Promise > Invalid props for component 'WarningDialog': 'message' is not a string`
> Screencast.from.06-03-2025.16.21.25.webm

Solution to this error inside other module of dependency in this PR:
https://github.com/OCA/account-fiscal-rule/pull/488

xaviedoanhduy:
> Solution to this error inside other module of dependency in this PR: #488

it's great to see it's resolved

account_avatax_exemption/controller/main.py
xaviedoanhduy: just a consideration. declare a new variable for example because in reality a lot of pagination has changed this value to 8, 12, 20,.. like the following example: https://github.com/odoo/odoo/blob/f60188fc9d2efc9e961fb4fecad800e529bee041/addons/website_hr_recruitment/controllers/main.py#L18

or we can keep the default `30` ([see](https://github.com/odoo/odoo/blob/f60188fc9d2efc9e961fb4fecad800e529bee041/addons/portal/controllers/portal.py#L17)) available in the function instead of overriding that value

account_avatax_exemption/views/partner_view.xml
xaviedoanhduy: thanks for clarifying, but isn't the purpose of this attribute to be hidden when the `parent_id` value is set? and its negation will include both `None` and `False` values.

version `14.0`: `attrs="{'invisible': [('is_company','=',False),('parent_id','!=',False)]}"`
verison `18.0`: `invisible="not is_company and parent_id"`

see: https://github.com/odoo/odoo/blob/f60188fc9d2efc9e961fb4fecad800e529bee041/addons/product/views/res_partner_views.xml#L9

account_avatax_exemption/models/exemption.py
```diff
@@ -0,0 +1,332 @@
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError, ValidationError
+
+
+class ExemptionRule(models.Model):
+    _name = "exemption.code.rule"
+    _description = "Avatax Custom Rules"
+
+    name = fields.Char(index=True, default=lambda self: self.env._("New"))
```
xaviedoanhduy:
```suggestion
    name = fields.Char(index=True, default=lambda self: _("New"))
```
We should not translate this value. so please restore it

account_avatax_exemption/models/exemption.py
```diff
@@ -0,0 +1,332 @@
...
+            if record.avatax_rate < 0 or record.avatax_rate > 100:
+                raise ValidationError(self.env._("Avatax rate range is from 0 to 100"))
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get("name", self.env._("New")) == self.env._("New"):
+                vals["name"] = self.env["ir.sequence"].next_by_code(
+                    "exemption.line.sequence"
+                ) or self.env._("New")
```
xaviedoanhduy:
```suggestion
            if vals.get("name", _("New")) == _("New"):
                vals["name"] = self.env["ir.sequence"].next_by_code(
                    "exemption.line.sequence"
                ) or _("New")
```

account_avatax_exemption/models/exemption.py
xaviedoanhduy:
```suggestion
    name = fields.Char(index="trigram", default=lambda self: self.env._("New"))
```
change `index=True` to `index='gin'` in: https://github.com/odoo/odoo/pull/83015/commits/12c655b497fbe101e94c9ecb68ab0b56c287d0f8#diff-695321b9df7129604a59fbc9bff34f97ba90b3394aa1166ca24e36b27389ae82

then from `index='gin'` to `index='trigram'` in: https://github.com/odoo/odoo/commit/a1904aa6f62acbd3d8e7d4cc873692e20edb37fd#diff-695321b9df7129604a59fbc9bff34f97ba90b3394aa1166ca24e36b27389ae82

account_avatax_exemption/views/avalara_salestax_view.xml
```diff
@@ -0,0 +1,110 @@
...
+        <field name="name">exemption.code.tree.view</field>
+        <field name="model">exemption.code</field>
+        <field name="type">list</field>
+        <field name="inherit_id" ref="account_avatax_oca.exemption_code_tree_view" />
+        <field name="arch" type="xml">
+            <xpath expr="//list" position="attributes">
+                <attribute name="string">Entity Use Code</attribute>
+            </xpath>
+        </field>
+    </record>
```
xaviedoanhduy:
```suggestion
```
see: https://github.com/OCA/pylint-odoo/issues/146

xaviedoanhduy (changes requested):

account_avatax_exemption/models/exemption.py
dnplkndll: is the _ fn something different in this context?

kobros-tech: putting _ here is translating or what?
@dnplkndll 
@xaviedoanhduy 
Do you need to translate or not translate?

xaviedoanhduy: > replace the call to translate a string (`_`) with `self.env._` for getting some performance improvement in some cases. See https://github.com/odoo/odoo/pull/174844.

only in some cases, also not needed. I checked the codebase and found that most of them will not translate this part

xaviedoanhduy (approved):
Looks good,

If possible, merge manual changes into a single commit `[MIG] $module: Migration to 18.0`

kobros-tech:
@dreispt 

Thanks, that we got the dependency base merged successfully.
This one I clean out of test requirements and its commit log has just the [MIG] commit.

dreispt (approved):

PR #479, opened 2025-01-22
[17.0][FIX] account_avatax[_sale]_oca: present theoretical tax rate on invoices

dreispt:
Tax names used to represent taxes are changed, to show both the theoretical tax (e.g. "9.5%") and the real tax rate needed to Odoo to calculate the exact same cent value (e.g. "9.576%").
And on Invoiced we, want only the theoretical tax to be presented (e.g. "9.5%").
This was also a chance to refactor of the code related to Avalara tax code creation, that was reduced by removing redundancies.

Sample output for the the new Avalara tax naming approach:

image

- [FIX] account_avatax_oca: present theoretical tax rate on invoices
- [FIX] account_avatax_sale_oca: present theoretical tax rate on invoices

### DIFF
--- a/account_avatax_oca/models/account_move.py
+++ b/account_avatax_oca/models/account_move.py
@@ -2,7 +2,6 @@ import logging
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError
-from odoo.tools import float_compare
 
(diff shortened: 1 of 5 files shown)


dreispt:
Needs review!

PR #484, opened 2025-02-06
[IMP] account_avatax_exemption_base: remove chatter from wizard view

kobros-tech:
chatter in wizard view does not make sense, it caused an issue in 17.0 and 18.0 migrations.
despite we don't see the side effect in 16.0!


dnplkndll (approved):
this is not needed here

wlin-kencove (approved):

dreispt (approved):

PR #488, opened 2025-03-08
[FIX] account_avatax_oca: return correct message format

kobros-tech:
the issue is mentioned in this message:

https://github.com/OCA/account-fiscal-rule/pull/475#issuecomment-2703271109

### DIFF
--- a/account_avatax_oca/models/partner.py
+++ b/account_avatax_oca/models/partner.py
@@ -21,15 +21,13 @@ class ResPartner(models.Model):
     @api.onchange("property_exemption_country_wide")
     def _onchange_property_exemption_contry_wide(self):
         if self.property_exemption_country_wide:
-            message = (
-                self.env._(
-                    "Enabling the exemption status for all states"
-                    " may have tax compliance risks,"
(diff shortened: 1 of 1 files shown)


account_avatax_oca/models/partner.py
```diff
@@ -34,7 +34,7 @@ def _onchange_property_exemption_contry_wide(self):
             return {
                 "warning": {
                     "title": self.env._("Tax Compliance Risk"),
-                    "message": message,
+                    "message": message[0],
```
rvalyi: I think you should just remove the comma line 32 instead of adding [0] (it seems parentheses were there to enforce the max line length, but it shouldn't be a tuple)

rvalyi (changes requested):

account_avatax_oca/models/partner.py
kobros-tech: thanks, you are correct and it solved the problem too

rvalyi (approved):

dnplkndll (approved):

PR #491, opened 2025-03-13
[16.0][FW] Avatax: Use Odoo Tax option

AlexPForgeFlow:
FW: https://github.com/OCA/account-fiscal-rule/commit/4c56a80fdc65d96d101a2d60a91f113104eee798


JordiBForgeFlow:
@dreispt Are you good with this? Not sure if you have time to review

dreispt:
@JordiBForgeFlow I don't understand what it trying to be done here and the use case.

AlexPForgeFlow:
Hi @dreispt !
This MR it's a forward port from one commit from v14, that got lost in v16. 
Original MR in v14: https://github.com/OCA/account-fiscal-rule/pull/237
Basically, this option allows you to override taxes and upload odoo computed taxes to avalara instead of recomputing taxes from avalara.
Use case we face up for example is, we have an invoice that taxes were previously computed and we want to make a refund to the customer, through a credit note. Now customer has a valid tax exemption certificate (so if taxes are recomputed tax amount will be 0), when confirming the credit note taxes will be recomputed again and now customer will have no taxes to pay, so refund it's not correct. With this MR, we ensure that the taxes that were originally computed are conserved in the refund process.

xaviedoanhduy:
hi guys, [this line](https://github.com/ForgeFlow/account-fiscal-rule/blob/4d5f2140f469c813b246276c78a56803d68359b1/account_avatax_oca/views/account_move_view.xml#L65) should be `account_avatax_oca.access_avatax_override_group` instead `account_avatax.access_avatax_override_group `

PR #496, opened 2025-03-15
[17.0][FIX] account_avatax[_sale]_oca: on multicompany, consider document company instead of env company (bump minor)

dreispt:
Taxes are wrong if user with a non-Avatax active company calculates taxes for an invoice of an Avatax-company.
Fixes usages of `self.env.company` with the document's company. 
Simplifies the access to the Avatax configuration, and in the process removes the annoying warnings in the log.

Changes (commits to review here):
- FIX] account_avatax_oca: Install on multi-country
- [FIX] account_avatax_oca: on multicompany, consider document company instead of env company
- [FIX] account_avatax_sale_oca: on multicompany, consider document company instead of env company

Dependencies (included):
- [ ]  #479
- [x] https://github.com/OCA/account-fiscal-rule/pull/499

### DIFF
--- a/account_avatax_oca/data/avalara_salestax_data.xml
+++ b/account_avatax_oca/data/avalara_salestax_data.xml
@@ -7,6 +7,7 @@
     </record>
     <record id="avatax_tax_group" model="account.tax.group">
         <field name="name">AvaTax</field>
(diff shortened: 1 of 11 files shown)


dreispt:
Rebased. Needs reviews.

PR #501, opened 2025-03-17
[16][IMP] account_ecotax: Add ecotax only if it match shipping address country

florian-dacosta:


mourad-ehm (approved):

PaulGoubert (approved):
LGTM

account_ecotax/models/ecotax_line_product.py
```diff
@@ -32,6 +32,13 @@ class EcotaxLineProduct(models.Model):
         help="Ecotax Amount computed form Classification or forced ecotax amount",
         store=True,
     )
+    country_ids = fields.Many2many(
+        "res.country",
+        string="Countries",
+        related="classification_id.country_ids",
+        help="Ecotax will be applied when delivered in the listed countries (or all "
```
hparfr: is the `help=` needed when there is a related field ?

florian-dacosta: No, it is not. I did not know it would somehow be taken from its main field, we don't need to duplicate the attribute !
I removed it, thanks.

hparfr (approved):
LGTM

dreispt (approved):

PR #502, opened 2025-03-17
[17][IMP] Store all fields in account move line ecotax model

florian-dacosta:
Forward port of https://github.com/OCA/account-fiscal-rule/pull/495


mourad-ehm (approved):

florian-dacosta:
fast-tracking as it is an easy fwd with no impact for users
/ocabot merge patch

PR #508, opened 2025-03-19
[16] account_ecotax and account_ecotax_sale  init computed fields on installation

florian-dacosta:
Creating these computed fields could be an issue for large database, so we create it and initialize it to 0.0 to avoid this issue.


Kev-Roche (approved):
LGTM

mourad-ehm (approved):

PR #510, opened 2025-03-21
[IMP] account_product_fiscal_classification: update views by python code.

legalsylvain:
**rational:**
In odoo, some module (like hr_expense) redefines product views. For each case, we need a glue module to insert fiscal_classification_id in the view. With that changes, insertion is done for all tree and form view.
Closes: #503, #504, #505

Also, fix : Display field also for product.produt form view.


legalsylvain (commented):
/ocabot merge minor

PR #511, opened 2025-03-26
[18.0][ADD] account_avatax_demo_oca

natuan9:


dreispt:
Demo data is usually included in the modules.
Is there a good reason to deviate from this and create a separate module?

nilshamerlinck:
beyond the demo data, the idea is to have a single module to introduce a new dev/user to the (complex) Avatax topic:
- installing the whole `*avatax*`  modules stack through the deps
- mocking the Avatax API not only in tests, so that an Avatax account would not be necessary

but it's still a wip for now

dreispt:
It will be easier to direct changes on each of the modules.
That will save you from the effort later of copy-pasting the changes to the right places.

For the related module discoverability problem, the standard practice if to have them as checkbox option in the Settings view.

Finally, in case you are interested in this: the current way the Avatax API is configured is not great,
Ideally it would be a section in the Invoicing/Accounting Settings, with values per company.
That can be a bit of a change and I haven't found budget to accommodate it yet.

dreispt:
@nilshamerlinck You mean to be able to download everything at once from the App store? Perhaps a `account_avatax_full_oca` module with just the hard dependencies?

For the tests with mock services, they are very welcome, but can go directly in the modules that are already there.

Issue #453, opened 2024-11-05
Split of the ecotax modules


florian-dacosta:
The ecotax modules did change a lot these last months. There are different ways to implement it.
1) The original module (`l10n_fr_account_ecotaxe`), where we do NOT use odoo taxes, because the ecotax is not actually a tax or at least a different type of tax because as far as I know it must count in the turnover.

2) The new module (`account_ecotax`) which DO use odoo taxes, because, it is quite practical.

From what I see, is that, depending on the use cases, the first solution may be preferable and for others, the second one is better. (At least, I have a customer, working on previous version who won't accept to migrate to the second one, with tax, I'll explain why below)

After looking at the differences between the version 1) and 2), what I see is that most of the code (maybe 90% or more) is the same.
So, I was wondering if it would be worth it to maintain both mode, or at least, split the present modules to allow the other implementation, even if we do not maintain the other implementation in the OCA.

As a first step, I propose to have : 
1 base module : `account_ecotax_base`, with 90% of the current code of current `account_ecotax`, basically every thing but what is linked to the `account.tax`
Having a second module `account_ecotax_tax` (or something like that, feel free to propose a better name) that would only implement the `account.tax` related features.

And do the exact same for the sale part : 
`account_ecotax_sale_base` and `account_ecotax_sale_tax`

Then, We'd be able to have 2 concurrent modules to `account_ecotax_tax` and `account_ecotax_sale_tax`, that would do the implementation without using the `account.tax`.
These 2 could not be in the OCA if we consider it would be to confusing to have 2 concurrent modules (which to choose in which cas maybe quite hard if you did not dig the subhect), we could decide that later.

The advantage is that all of 98% of the database structure needed for ecotax modules would be common, so in the base modules.
Additional modules helping the declaration, doing statistics of whatever should also work for both cases (could depend only on the base modules, because it would be using the common model `account.move.line.ecotax`)

Now, I'll try to list a bit the advantages of both modes.
Here is the list of differences I see between the 2 ways : 

Advantages using odoo's tax system : 
1) You can choose if the ecotax amount is included in the price_unit (by setting the tax as included or not)
It is a nice feature, mainly for B2B companies or at least some of them.
2) You can choose if you want to apply the discount on the price whithout ecotax or with ecotax (by setting the same parameter as above, if price is not included, discount won't apply on the ecotax amount).
[...]

PR #513, opened 2025-03-28
[16][ADD] account_ecotax_accounting

florian-dacosta:
This module allow to create a misc accounting entry to reclassify the ecotax amounts and isolate it in a dedicated account.
Indeed, using the base ecotax module (`account_ecotax`), the ecotax amounts are including in the revenue/product accounts and usually you would prefer have it isolated.

Note that the module `account_ecotax_tax` that uses odoo taxes to manage ecotax, already has the feature of isolating the ecotax amounts using the native tax features. But like explained here : https://github.com/OCA/account-fiscal-rule/issues/453 there are pros and cons for this last module and in some cases, it is better to use only the first module (`account_ecotax`) but still beeing able to isolate the ecotax amounts.

This is the problem that this module tries to solve.

Important note : this module is NOT needed nor wanted when using `account_ecotax_tax`, like I just explained, so I put an exclusion in the manifest.


hparfr:
There is an issue with test exclusion

florian-dacosta:
> There is an issue with test exclusion

Yes I know, but I have no idea on how to get around this.
The CI install all modules of the repo, I do not know if there is a way to make it different or not.

hparfr (approved):
LGTM

PR #514, opened 2025-03-28
[17][account_ecotax,account_ecotax_sale] Add ecotax only if it match shipping address country

florian-dacosta:
Forward port of https://github.com/OCA/account-fiscal-rule/pull/501

This does not change the current behavior when installed because if you do not configure any country on the classification, it still applies for every country.

Maybe it could interest you @vvrossem @SilvioC2C


mourad-ehm (approved):

florian-dacosta:
Well, it seems to work well from UI, but I am not sure why the tests are failling... The creation of the ecotax line is not done correctly from the test.

PR #516, opened 2025-04-04
[16.0] mig account_fiscal_position_tax_included

legalsylvain:
This module extend the Odoo account module, regarding fiscal position.

By default, it is only possible to realize such tax mapping :

* Tax Excl --> Tax Excl / No tax
* Tax Incl --> Tax Excl / No tax
* Tax Incl --> Tax Incl / No tax

This module allows to realize such mapping and fixes incorrect computation.

* Tax Excl --> Tax Incl


quentinDupont:
Internal message : MeP 27/07/2025 006-asperges-vertes

PR #489, opened 2025-03-12
[17.0][IMP] l10n_eu_oss_oca: Czech Republic, Estonia, Finland VAT

manuelregidor:
Change in VAT in Czech Republic, Estonia, Finland

@HaraldPanten @ValentinVinagre @Jesarregui 

T-7852

PR #519, opened 2025-04-07
[15.0][IMP] l10n_eu_oss_oca: Czech Republic, Estonia, Finland VAT

BetoRaja:
Change in VAT in Czech Republic, Estonia, Finland

Updated values based on data from: https://github.com/OCA/account-fiscal-rule/pull/489

@HaraldPanten @ValentinVinagre @luis-ron 

T-7902


ValentinVinagre (changes requested):
why this imp have 3 commits?

PR #522, opened 2025-04-09
[17.0][FIX] account_avatax_sale[_oca]: on multicompany, consider document company instead of env company

dreispt:
Depends on (and includes) #479

- [FIX] account_avatax_oca: on multicompany, consider document company instead of env company
- [FIX] account_avatax_sale_oca: on multicompany, consider document company instead of env company


dreispt:
Already included in #496

PR #371, opened 2023-08-03
[16.0][FIX] account_avatax_oca: tax calculation improvements

RLeeGML:
Improvements to v16 migration

Calculates taxes properly for lines that have split taxes. E.g. in Tennessee taxes, one line for $100 may have two applicable taxes, 4% on $75 and 4.5% on $25. Odoo should consider the total tax amount for that line ($4.13) and compute the tax rate (4.125%) based on that.

Bugfixes for unbalanced entry error, changed core Odoo method to _sync_dynamic_lines to recompute journal items

PR #529, opened 2025-05-14
[16.0][IMP] account_avatax_[sale]_oca: support multiple taxes appearing on each line instead of one

xaviedoanhduy:
**Context:**

 The issue addressed in [PR](https://github.com/OCA/account-fiscal-rule/pull/371) involves handling split tax lines from Avalara—where a single invoice line (e.g., \$100) might be taxed at multiple rates (e.g., 4% on \$75 and 4.5% on \$25). however recalculating the rate leads to incorrect rate values as the amounts involved in the recalculation have been rounded by Avalara.

**What this PR improves:** 

This PR enhances the solution with a cleaner and more maintainable approach:

* **Uses Avalara API `details` directly:** Instead of calculating an average rate or recaculate rate, this PR reads each component tax (rate, taxName) directly from the API response.
* **Creates one Odoo tax per detail:** Each tax line in the API response maps to a specific `account.tax` record in Odoo, improving traceability and avoiding ambiguity.
* **Prevents unnecessary tax creation:** Ensures tax records are generated with valid, legit names (with absolute rate) (e.g., `Local Tax [Code] 4.5%`) and prevents issues caused by auto-generating taxes with obscure rate names (e.g., `9.75**`).
* **More transparent tax totals:** Makes the resulting tax structure more aligned with Avalara’s breakdown and more user-friendly - This allows us to easily compare tax information on each line from the API response logs.

**Example from PR results and a little review of some issues**: We should pay attention to the first line

- Before PR-371 merged: if we sum all percent -> the total percentage will be <=> 7 + 2.75 + 2.25 + 0.5 = 12.5% => total taxes = 750,125 instead of 585,10
- Current state of code (after PR#371): the current code has recalculated the correct tax value for each small detail, but there are some issues with the recalculated rate value not being an absolute and valid number for tax as mentioned above.
- With this PR: Each individual rate is preserved as-is and generates clean tax lines for each line, such as image below

image

### DIFF
--- a/account_avatax_oca/hooks.py
+++ b/account_avatax_oca/hooks.py
@@ -100,3 +100,31 @@ def post_load_hook():  # noqa: C901
     if not hasattr(AccountMoveLine, "_compute_all_tax_origin"):
         AccountMoveLine._compute_all_tax_origin = AccountMoveLine._compute_all_tax
     AccountMoveLine._compute_all_tax = _compute_all_tax_new
+
+    @api.depends("quantity", "discount", "price_unit", "tax_ids", "currency_id")
+    def _compute_totals_new(self):
+        for line in self:
+            if line.display_type != "product":
+                line.price_total = line.price_subtotal = False
+            # Compute 'price_subtotal'.
+            line_discount_price_unit = line.price_unit * (1 - (line.discount / 100.0))
+            subtotal = line.quantity * line_discount_price_unit
+
+            # Compute 'price_total'.
+            if line.tax_ids:
+                taxes_res = line.tax_ids.with_context(current_aml=line.id).compute_all(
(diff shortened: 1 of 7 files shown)


dreispt:
@xaviedoanhduy Originally I had concerns this could lead to more rounding issues.
Looking back from a distance, it might be the case this actually helps reducing rounding differences.

I'm in to taking this, and even have it enabled by default.
It should be an option, to not disturb existing users that don't want to change.

xaviedoanhduy:
The red CI is not coming from this module, but from `account_ecotax` itself on the `16.0` branch, see: https://github.com/xaviedoanhduy/account-fiscal-rule/pull/7

PR #539, opened 2025-07-11
[18.0] [MIG] account_fiscal_position_autodetect_optional_vies: Migration to 18.0

bizzappdev:


rafaelbn:
/ocabot migration account_fiscal_position_autodetect_optional_vies

rafaelbn:
Checking if runboat works

rafaelbn:
Hello @bizzappdev ,

I Think this module is not needed any more

https://www.odoo.com/documentation/18.0/applications/finance/accounting/taxes/vat_verification.html

@pedrobaeza @victoralmau you are authors. What do you think?

Regards

pedrobaeza:
But is there any option in the fiscal position to select it automatically depending on the VIES?

loida-vm:
> But is there any option in the fiscal position to select it automatically depending on the VIES?

Yes, it does
I have tested it in version 16, and it recalculates the fiscal position according to the shipping address (country), as mentioned in the README.
If the fiscal position has the checkbox selected:
imagen

Then it is recalculated when creating a sale or invoice.
imagen

I have done the same test in version 18, and the procedure is exactly the same:
imagen

It calculates based on the shipping address:
imagen

rafaelbn:
@bizzappdev please rebase we need runboat to test it, thanks!

pedrobaeza:
VIES is for intra-community, not extra-community, and this module introduces the check `vat_vies_required`, that is the one doing the difference.

rafaelbn:
@pedrobaeza I've tested in runboat here in OCA and this module is not needed any more. Why?

Partner 1: Intra
Partner 2: Non intra

Now Odoo have field in res.partner that check when invoicing for selecting the proper FP

image

So it works out-of-the-box

Closing! 😄

PR #545, opened 2025-08-01
[16] Fix build fixing account_ecotax_tax tests

florian-dacosta:
Since https://github.com/odoo/odoo/commit/13e9833e0bc809a26843890363586f61a37d061c
it seems that having a python code tax with included price option along with another included price option tax that take the first tax into account does not work anymore.

In order to make the CI green again, I did change the test of the module account_tax_ecotax in order to test with price excluded taxes and avoid the issue, at least in the tests.
From what I saw, we do not need this account_ecotax_tax module for the price included tax case, the module is mainly usefull to have price excluded ecotax management.
Also the but does not seem to happen in Odoo 18, so I think the problematic case won't really be an issue in v16 and won't be for further versions.

This PR does NOT need to be ported in following versions nor need to be taken into account in the migration of this module. (although it is not really an issue if it is)

### DIFF
--- a/account_ecotax/tests/test_ecotax.py
+++ b/account_ecotax/tests/test_ecotax.py
@@ -203,8 +203,8 @@ class TestInvoiceEcotaxCommon(TransactionCase):
         return new_qtys
 
     def _run_checks(self, inv, inv_expected_amounts, inv_lines_expected_amounts):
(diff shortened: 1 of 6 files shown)


florian-dacosta:
I'll fast-track it as it fixes the CI on v16 + it can't add new regression since only test files have changed.
But remarks or suggestion are welcome to make further change afterward if needed.

PR #547, opened 2025-08-12
[15.0][IMP] l10n_eu_oss_oca: estonia, romania

ValentinVinagre:
Change VAT in Estonia and Romania

@HaraldPanten  @luis-ron 

T-8779


HaraldPanten (approved):
LGTM as VAT has been increased from 01/07/2025 in Estonia, and from 01/08/2025 in Romania.

@ValentinVinagre We should port this changes to other versions.

/ocabot merge patch

PR #551, opened 2025-09-04
[14.0] [FIX] account_product_fiscal_classification: Multi company fixes

paradoxxxzero:
The `ir.rule` is wrong and should use `company_ids` (which is the case in the 16.0 version). 

Also a classification without company should be instance wide.


legalsylvain (approved):
makes sense.

legalsylvain (commented):
/ocabot merge minor

PR #552, opened 2025-09-05
[18.0][ADD] account_fiscal_position_vies_warning

Andrii9090-tecnativa:
This module was developed to prevent errors when setting the Fiscal Position to Intra-community if the partner’s VAT number is not registered in VIES.

When the partner's VAT number is not registered in VIES and the Fiscal Position is set to Intra-community, a ValidationError is raised. This check works in the `res.partner `and `account.move models`.

It's short video with fucntional tests 
https://www.loom.com/share/b2257ce8fc484f2c8e41fcc1406f0fac?sid=c09b748b-49e2-4738-a7a6-e2ca9f07419d

@rafaelbn @yajo @chienandalu @Shide can you review, please?
MT-11216 @moduon


account_fiscal_position_vies_warning/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+
+from odoo import _, models
+from odoo.exceptions import ValidationError
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def action_post(self):
+        if self.env.company.vat_check_vies:  # Verify VAT Numbers is True
```
chienandalu: - Early returns tend to be more readable
- Use comments when they complement information :)
```suggestion
        if not self.env.company.vat_check_vies: 
            return super().action_post()
```

account_fiscal_position_vies_warning/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+            company_id = self.env.company.id
+            try:
+                if (
+                    self.fiscal_position_id
```
chienandalu: You're assuming a single record logic, but the main method works over recordsets. And thinking about this, if different companies invoices are posted in the method (maybe some purchase-sale intercompany logic), you should be looking the `vat_check_vies` check for everyone, not just the active one...

account_fiscal_position_vies_warning/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    @api.constrains("property_account_position_id")
+    def check_fiscal_position_and_vies_valid(self):
+        if not self.env.company.vat_check_vies:  # Verify VAT Numbers is True
```
chienandalu: Code is self explanatory in this case
```suggestion
        if not self.env.company.vat_check_vies:
```

account_fiscal_position_vies_warning/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+                    == self.env.ref(f"account.{company_id}_fp_intra")
+                    and not self.partner_id.vies_valid
+                ):
```
chienandalu: Mmm... well, this is very dependant on `l10n_es` xml_ids, so it's very rigid IMO.

Maybe a more generic way would be to have a set of fiscal positions at company level and check them here

account_fiscal_position_vies_warning/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    def action_post(self):
+        if self.env.company.vat_check_vies:  # Verify VAT Numbers is True
+            company_id = self.env.company.id
+            try:
+                if (
+                    self.fiscal_position_id
+                    == self.env.ref(f"account.{company_id}_fp_intra")
```
chienandalu: Not the right approach as said, but in any case you wouldn't need the `try` if you used `raise_if_not_found`
```suggestion
                    == self.env.ref(f"account.{company_id}_fp_intra", raise_if_not_found=False)
```

account_fiscal_position_vies_warning/models/res_partner.py
```diff
@@ -0,0 +1,31 @@
...
+                ):
+                    raise ValidationError(
+                        _(
+                            "You can't set the Fiscal Position to Intra-community, "
+                            "[%s] doesn't pass VIES VAT number validation!"
+                        )
+                        % record.name
+                    )
+        except ValueError:
+            return
```
chienandalu: The same comments as in `account.move` apply

account_fiscal_position_vies_warning/tests/common.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2025 Moduon Team S.L. <info@moduon.team>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.tests import TransactionCase
+
+
+class TestViesWarningCommon(TransactionCase):
```
chienandalu: Use BaseCommon as base test class to reduce overhead from tracking and boost the test suite.

```suggestion
class TestViesWarningCommon(BaseCommon):
```

account_fiscal_position_vies_warning/tests/test_res_partner_vies_warning.py
```diff
@@ -0,0 +1,42 @@
...
+from odoo.exceptions import ValidationError
+from odoo.tests import Form
+
+from odoo.addons.account_fiscal_position_vies_warning.tests.common import (
+    TestViesWarningCommon,
+)
+
+
+class TestViesWarningResPartner(TestViesWarningCommon):
+    def test_set_fiscal_position_intra_community_for_partner_not_vies(self):
```
chienandalu: Do you really need two separate test case classes?

chienandalu (changes requested):
Thanks @Andrii9090 ! Some change suggestions below:

account_fiscal_position_vies_warning/models/account_move.py
Andrii9090-tecnativa: Hi @chienandalu! I updated this module and added a new field `check_vies_fiscal_position_ids` to the `res.company` model. This makes the module more flexible.

Thank you!

account_fiscal_position_vies_warning/models/account_move.py
```diff
@@ -0,0 +1,30 @@
...
+                raise ValidationError(
+                    _(
+                        "You can't set the Fiscal Position to %(fiscal_pos)s, "
+                        "%(partner_name)s doesn't pass VIES VAT number validation!"
+                    )
+                    % {
+                        "fiscal_pos": record.fiscal_position_id.name,
+                        "partner_name": self.partner_id.name,
+                    }
+                )
```
chienandalu: A method at partner's level for raising this message would avoid the need for translating it twice

account_fiscal_position_vies_warning/tests/common.py
chienandalu, line 1: What I meant in my last revision was to do all the testing in a single class. No need to have separate classes and a common ground for such a small feature :)

chienandalu (changes requested):
Looking very goood now :)

Some additional comment (mostly formal)

Is ready for functional testing in any case

Gelojr (approved):
Great work on this contribution @Andrii9090 !
We performed the following tests:

- Test 1: Tried to save a partner without VIES validation using a fiscal position listed in Company “Check Vies Fiscal Position”; Odoo raised the warning and the record could not be saved (OK).
- Test 2: Saved a partner with a VIES-valid VAT using a protected intra-community fiscal position; the partner was saved successfully (OK).
- Test 3: Saved a partner without VIES validation using a non-protected fiscal position; the partner was saved successfully (OK).
- Test 4: Created an invoice for a non-VIES partner with a protected fiscal position and attempted to post; posting was blocked with the warning (OK).
- Test 5: Created an invoice for a VIES-valid partner with a protected fiscal position and posted; the invoice was posted successfully (OK).
- Test 6: Removed the intra-community fiscal position from the Company watchlist and retried saving a non-VIES partner with that position; saving was allowed (OK).
- Test 7: Disabled “Verify VAT Numbers” and retried saving/posting with protected positions; no warnings appeared and both save and post proceeded (OK).

account_fiscal_position_vies_warning/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def raise_vies_warning(self, partner_name, fiscal_position_name):
+        raise ValidationError(
+            _(
+                "You can't set the Fiscal Position to %(fiscal_pos)s, "
+                "%(partner_name)s doesn't pass VIES VAT number validation!"
+            )
+            % {
+                "fiscal_pos": fiscal_position_name,
+                "partner_name": partner_name,
```
chienandalu: You can use `self` and take the data from the record:
```suggestion
    def raise_vies_warning(self, fiscal_position_id):
        raise ValidationError(
            _(
                "You can't set the Fiscal Position to %(fiscal_pos)s, "
                "%(partner_name)s doesn't pass VIES VAT number validation!"
            )
            % {
                "fiscal_pos": fiscal_position_id.name,
                "partner_name": self.name,
```

account_fiscal_position_vies_warning/models/res_partner.py
```diff
@@ -0,0 +1,35 @@
...
+            return
+        for record in self:
+            if (
+                record.property_account_position_id
+                in self.env.company.check_vies_fiscal_position_ids
+                and not record.vies_valid
+            ):
+                self.raise_vies_warning(
+                    record.name, record.property_account_position_id.name
+                )
```
chienandalu: And then:
```suggestion
                self.raise_vies_warning(record.property_account_position_id)
```

account_fiscal_position_vies_warning/models/account_move.py
Andrii9090-tecnativa: Now it has only one method in `res.parnter`
Thank you!

account_fiscal_position_vies_warning/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+            return super().action_post()
+        for record in self:
+            if (
+                record.fiscal_position_id
+                in self.env.company.check_vies_fiscal_position_ids
+                and not record.partner_id.vies_valid
+            ):
+                self.partner_id.raise_vies_warning(
+                    self.partner_id.name, record.fiscal_position_id.name
+                )
```
chienandalu: Careful here, `self` is the recordset. You want to call the method over the current `record` in your loop
```suggestion
                record.partner_id.raise_vies_warning(record.fiscal_position_id)
```

chienandalu (changes requested):

account_fiscal_position_vies_warning/models/res_partner.py
Shide:
```suggestion
                record.raise_vies_warning(record.property_account_position_id)
```

account_fiscal_position_vies_warning/models/account_move.py
Andrii9090-tecnativa: Thanks, that’s a great idea :+1:  I’ve updated this part of the code

chienandalu (approved):
Nice! (just code review)

rafaelbn (changes requested):
Hello! thank you for this improvement 😃 

After reviewing I think we should change the way we trigger this warning

- Configuración should be at fiscal position level
- Activation (trigger) when you set the fiscal position with this configuration.

In a international software with many countries outside EU we should prevent add configurations in the res.company form that only affect in some cases.

Does my argument seem reasonable to you? 🙏🏼 

Thank you! ❤️

Andrii9090-tecnativa:
@rafaelbn it's a nice idea set configuration in the fiscal position level!
I changed the code of this module.
@Gelojr @chienandalu @Shide can you review?

This is a short video with functional test
https://www.loom.com/share/e6cf0b6f9f1f441b9c7a7ab46f366a99?sid=6d30a70d-f2a0-444a-b425-d2d5316f5d3d

loida-vm (approved):
Functional review done!
The VAT format is first checked and then validated via VIES, showing a warning only if the fiscal position has 'Show Vies Warning' enabled. So, it works! 

But, I believe this warning should be enabled by default to prevent potential errors

Thank you @Andrii9090

Andrii9090-tecnativa:
@loida-vm thank you for your review :smile:
If this warning is enabled by default, as we know, the Fiscal positions that need to be checked?

Gelojr (approved):
Great work on this contribution @Andrii9090 !
We performed the following tests:

- Test 1: Tried to save a partner without VIES validation using a fiscal position with “Show Vies Warning”; Odoo raised the warning and the record could not be saved (OK).
- Test 2: Saved a partner with a VIES-valid VAT using a fiscal position with “Show Vies Warning”; the partner was saved successfully (OK).
- Test 3: Saved a partner without VIES validation using a fiscal position without “Show Vies Warning”; the partner was saved successfully (OK).
- Test 4: Created an invoice for a non-VIES partner with a fiscal position with “Show Vies Warning” and attempted to post; posting was blocked with the warning (OK).
- Test 5: Created an invoice for a VIES-valid partner with a fiscal position with “Show Vies Warning” and posted; the invoice was posted successfully (OK).
- Test 6: Disabled “Show Vies Warning” on the fiscal position and retried saving a non-VIES partner with that position; saving was allowed (OK).
- Test 7: Disabled “Verify VAT Numbers” in settings and retried saving/posting with a warned fiscal position; no warnings appeared and both save and post proceeded (OK).

Andrii9090-tecnativa:
@rafaelbn can you review?
Thanks :heart:

rafaelbn (approved):
Functional review 👍🏼 

Thank you!

/ocabot merge patch

PR #556, opened 2025-10-17
[17.0][FIX] l10n_eu_oss_oca: set country on tax groups

LauraCForgeFlow:
This fix sets `country_id` for all the OSS tax groups, upon creation as well as on the migration script. This avoids the `tax_group_id` from being changed and set incorrectly by [this method](https://github.com/odoo/odoo/blob/eee013201da79716fedaee5185e6dbebb046ab52/addons/account/models/account_tax.py#L222) on OSS taxes when changing the `country_id` in the migration script.

Without this change, when modifying the `country_id` of the OSS taxes, their tax group would be recomputed and it could be assigned to a wrong group (not even related to OSS).


JordiMForgeFlow (approved):
code LGTM

JasminSForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM

MiquelRForgeFlow (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #560, opened 2025-11-14
[16.0][ADD] account_avatax_oca_line_grouping

crrodrigueztrujillo:
@BinhexTeam

This PR adds a new addon: account_avatax_oca_line_grouping.

Some US jurisdictions (e.g. Florida, Rule 12A-15.004) allow or require that multiple items sold together as a single sale / working unit are taxed on the total invoice amount, with discretionary surtax caps applied globally instead of line by line.

The standard account_avatax_oca connector always sends one AvaTax line per Odoo line, which makes it difficult to align with these “single sale / working unit” rules while keeping AvaTax as the tax engine.

This addon introduces an optional, company-level setting “Group document lines for AvaTax” that:

Aggregates all lines of a Sales Order / Customer Invoice into a single AvaTax line when calling AvaTax.

Lets AvaTax compute tax on the total while Odoo still keeps the standard multi-line document.

Preserves the existing behaviour when the option is disabled (full backward compatibility).

This is particularly useful for industries (construction, materials, project-based sales) where the whole invoice must be treated as a single taxable unit.


jelenapoblet:
@crrodrigueztrujillo since this is specific for the US, it should go to l10n_us repo, can you move it please ?

PR #561, opened 2025-11-30
[17.0][MIG] account_ecotax_tax

rvalyi:
standard migration

PR #562, opened 2025-11-30
[18.0][MIG] account_ecotax_tax

rvalyi:
standard migration from the 17.0 migration #561


account_ecotax_tax/models/account_tax.py
```diff
@@ -22,11 +22,21 @@ def onchange_is_ecotax(self):
...
-            self.python_compute = """
+            self.formula = """
 # price_unit
 # product: product.product object or None
 # partner: res.partner object or None
-# for weight based ecotax
-# result = quantity and  product.weight_based_ecotax * quantity or 0.0
-result = quantity and product.fixed_ecotax  * quantity or 0.0
+# for weight based ecotax:
+# quantity and  product.weight_based_ecotax * quantity or 0.0
```
florian-dacosta: comments are not allowed anymore, I run into the erreor : 
```
Malformed formula '# price_unit
# product: product.product object or None
# partner: res.partner object or None
# for weight based ecotax:
# quantity and  product['weight_based_ecotax'] * quantity or 0.0
quantity and product['fixed_ecotax'] * quantity or 0.0' at position 0
```

I guess we have to remove the whole comment.
I am not sure if there is a reason to create 2 ecotax taxes, one for fixed amount an the other one for weight amount, like suggested by the comment.
Is there any reason not to used the `product.ecotax_amount` instead, to have the whole amount ?

Also, if we only have one excotax, tax, I guess it would make sense to create it by default ? Well this could be a latter improvement of course, not really related to the migration.

Otherwise, it seems to work as expected.

rvalyi: > comments are not allowed anymore, I run into the erreor :
> 
> ```
> Malformed formula '# price_unit
> # product: product.product object or None
> # partner: res.partner object or None
> # for weight based ecotax:
> # quantity and  product['weight_based_ecotax'] * quantity or 0.0
> quantity and product['fixed_ecotax'] * quantity or 0.0' at position 0
> ```

You are right, we also removed it for our other customer who tested it all with their accountant:



> 
> I guess we have to remove the whole comment. I am not sure if there is a reason to create 2 ecotax taxes, one for fixed amount an the other one for weight amount, like suggested by the comment. Is there any reason not to used the `product.ecotax_amount` instead, to have the whole amount ?

Look I didn't design the module and I'm not sure either about  that... (may be it came from C2C?) All I can tell you is that we got an accountant validated everything for France when configured as above.

> 
> Also, if we only have one excotax, tax, I guess it would make sense to create it by default ? Well this could be a latter improvement of course, not really related to the migration.

Done. Notice I had to hard-code it for country_id France to get tests pass. But it is also noupdate=True so I think this is de good default.

> 
> Otherwise, it seems to work as expected.

That's what we found as well...

account_ecotax_tax/models/account_tax.py
```diff
@@ -22,11 +22,16 @@ def onchange_is_ecotax(self):
...
             self.include_base_amount = True
-            self.python_compute = """
-# price_unit
-# product: product.product object or None
-# partner: res.partner object or None
-# for weight based ecotax
-# result = quantity and  product.weight_based_ecotax * quantity or 0.0
-result = quantity and product.fixed_ecotax  * quantity or 0.0
+            self.formula = """
+quantity and product.fixed_ecotax * quantity or 0.0
```
florian-dacosta, line 26: Well, since there is no obvious reason to have one dedicated tax to fixed ecotax and one for weight ecotax, I'd prefer to switch on one taxe for both case by default, seems simpler.
If it becomes an issue, we'll just  change back and add a comment somewhere to explain why it would not be advised.
In the meantime, one can still duplicated the tax manually to have both if he likes.
The only reason I see to add multiple ecotax tax is to manage multiple account.account. But, we could also want multiple fixed amount tax, it depends on the accountant I guess.

account_ecotax_tax/data/account_tax_data.xml
```diff
@@ -0,0 +1,51 @@
...
+    </record>
+
+    <record id="tax_ecotax" model="account.tax">
+        <field name="name">EcoTax</field>
+        <field name="type_tax_use">sale</field>
+        <field name="amount_type">code</field>
+        <field name="is_ecotax" eval="True" />
+        <field
+            name="formula"
+        >quantity and product.fixed_ecotax * quantity or 0.0</field>
```
florian-dacosta, line 14: Well, since there is no obvious reason to have one dedicated tax to fixed ecotax and one for weight ecotax, I'd prefer to switch on one taxe for both case by default, seems simpler.
If it becomes an issue, we'll just change back and add a comment somewhere to explain why it would not be advised.
In the meantime, one can still duplicated the tax manually to have both if he likes.
The only reason I see to add multiple ecotax tax is to manage multiple account.account. But, we could also want multiple fixed amount tax, it depends on the accountant I guess.

florian-dacosta (approved):
Minor remark I think it would be good  to change but not blocking.

> Done. Notice I had to hard-code it for country_id France to get tests pass. But it is also noupdate=True so I think this is de good default.

No sure about this one, not sure it is a good thing to make a change so the test pass, can't we adapt the test if needed.
Anyway, we may go this way and see how we adapt, I think that the ecotax management for foreign countries is not well managed in the module yet, we'll probably have to improve stuff around this.

hparfr:
@rvalyi what is the status of this PR ?

hparfr (approved):

PR #565, opened 2025-12-05
[ADD] account_tax_product_amount: set tax amount per product variant

petrus-v:
This module allows defining specific fixed tax amounts for each product variant. This is essential for use cases like Ecotax, where the tax amount depends on variant attributes (size, weight, etc.) rather than a single global value.

### Key Features:

- Adds a Use product amount option to Fixed tax types.
- Allows setting different tax amounts per variant in the product configuration.
- Compatible with Sales, Purchases, and Invoicing.

### How to test:

1. Create a Tax with Tax Computation = *Fixed* and check *Use product amount* (for ecotax, also check *Affect Base of Subsequent Taxes* so it is included in the VAT base).
2. Add this tax to a Product.
3. Open the Product Variants and set a specific amount for this tax on each variant.
4. Create a Sale Order or Invoice and verify that the tax amount matches the specific variant configuration.


petrus-v:
Hi @rvalyi,
I’m not fully aware of your specific requirements regarding eco-taxes, but could you consider this alternative approach?

My impression is that it could serve as a base module for all eco-tax implementations, while keeping the code footprint smaller.

I would be glad if you could share your thoughts on this approach. I’d also be interested to understand whether you have already evaluated something similar in the past, and if so, what led you to choose a different direction.

petrus-v:
I've just rebase on top of branch 18.0.

@gregory-moka @Rom10811 could you review this PR ?

PR #574, opened 2026-03-02
[18.0][FIX] account_avatax_oca: update Avalara certification code

dreispt:
Avalara requires the connector to go through re-certification.
This has been done by the OSI team, the new assigned client code should
be added.

Using a connector with a properly certified code is important to be able
to access Avalara's customer support if that is needed.


hbrunn (approved):
@dreispt is there any publicly available documentation about the new code? This looks fine technically, but I can't say if this code is correct

PR #586, opened 2026-04-16
[18][FP][account_ecotax] Store all fields in account move line ecotax model for further reporting

florian-dacosta:
Forward port of https://github.com/OCA/account-fiscal-rule/pull/495

@rvalyi @mourad-ehm @bealdav


bealdav (approved):
Thanks a lot

mourad-ehm (approved):

PR #589, opened 2026-05-07
[19.0][MIG] account_ecotax: Migration to 19.0

rvalyi:
Migration of account_ecotax from 18.0 to 19.0.


account_ecotax/models/product_product.py
```diff
@@ -0,0 +1,95 @@
...
+    def _compute_all_ecotax_line_product_ids(self):
+        for product in self:
+            product.all_ecotax_line_product_ids = (
+                product.ecotax_line_product_ids
+                | product.additional_ecotax_line_product_ids
+            )
+
+    def _search_all_ecotax_line_product_ids(self, operator, operand):
+        if operator in Domain.NEGATIVE_OPERATORS and operand:
+            return [
```
hbrunn, line 58: this should be `Domain` objects

hbrunn (changes requested):

account_ecotax/models/account_move.py
```diff
@@ -0,0 +1,65 @@
...
+        compute="_compute_ecotax",
+    )
+
+    @api.depends("invoice_line_ids.subtotal_ecotax")
+    def _compute_ecotax(self):
+        for move in self:
+            move.amount_ecotax = sum(move.line_ids.mapped("subtotal_ecotax"))
+
+    @api.model
+    def _get_tax_totals(
```
vvrossem, line 25: IIRC, _get_tax_totals is a 15.0 method: https://github.com/odoo/odoo/blob/15.0/addons/account/models/account_move.py#L1874 and doesn't exist anymore since 16.0

Camille0907:
Superseded by https://github.com/OCA/account-fiscal-rule/pull/606 to take review comments into account and add a French translation

PR #601, opened 2026-06-03
[18.0][MIG] account_fiscal_position_type

hbrunn:


account_fiscal_position_type/models/account_fiscal_position.py
```diff
@@ -0,0 +1,21 @@
...
+
+    _TYPE_POSITION_USE_SELECTION = [
+        ("sale", "Sale"),
+        ("purchase", "Purchase"),
+        ("all", "All"),
+    ]
+
+    type_position_use = fields.Selection(
+        string="Position Application",
+        selection=_TYPE_POSITION_USE_SELECTION,
```
StefanRijnhart: Not sure if you agree, but delegating the selection to a class member looks overengineered to me now that it is not reused in other classes.

StefanRijnhart (approved):
Thanks!

account_fiscal_position_type/models/account_fiscal_position.py
hbrunn: I agree very much, thanks!

StefanRijnhart: Thanks for the update!

PR #603, opened 2026-06-25
[19.0] [MIG] account_multi_vat

ThomasBinsfeld:
From #386 
Refs #555


account_multi_vat/models/account_move.py
```diff
@@ -0,0 +1,31 @@
...
+        string="Customer tax administration",
+        comodel_name="res.partner",
+        ondelete="restrict",
+        domain=[("is_tax_administration", "=", True)],
+    )
+
+    @api.depends("partner_id", "customer_vat_partner_id")
+    def _compute_customer_vat(self):
+        for rec in self:
+            if not rec.partner_id or rec.move_type not in ("out_invoice", "out_refund"):
```
acsonefho: The best is probably to use `rec.is_sale_document()` ?

acsonefho (changes requested):

account_multi_vat/models/account_move.py
ThomasBinsfeld: Done

acsonefho (approved):

PR #605, opened 2026-07-29
[16.0][FIX] account_avatax_oca, account_avatax_sale_oca: support tax included price

almumu:
### **Problem**

The price included tax does not come from US tax rules: the setup is a global shop whose standard is tax inclusive pricing, integrated with Odoo, so the order line carries a gross price_unit with a price included tax while AvaTax is what determines the tax.

Taxes flagged as price_include are embedded in price_unit, so the raw line amount is gross. Both line preparers report that gross amount as the AvaTax taxable base:

  ```python
  # account_avatax_oca/models/account_move.py, _get_avatax_amount
  base_amount = base_line.price_unit * quantity

  # account_avatax_sale_oca/models/sale_order.py, _avatax_prepare_line
  amount = sign * line.price_unit * line.product_uom_qty * (1 - line.discount / 100.0)
  ```

The AvaTax LineItemModel.amount field is documented as "Total amount for this line", and a price included tax is not part of that taxable base. The result is that AvaTax records a  base and a tax that are both overstated, while the Odoo document itself stays balanced — which is what makes the defect hard to notice from Odoo alone.

Worked example, 8.75% tax with price_include=True on a gross unit price of 340:

<pre>  ┌─────────────────────────────┬──────────────┬───────┬────────┐
  │                             │ Taxable base │  Tax  │ Total  │
  ├─────────────────────────────┼──────────────┼───────┼────────┤
  │ Odoo document               │ 312.64       │ 27.36 │ 340.00 │
  ├─────────────────────────────┼──────────────┼───────┼────────┤
  │ Reported to AvaTax (before) │ 340.00       │ 29.75 │ 369.75 │
  ├─────────────────────────────┼──────────────┼───────┼────────┤
  │ Reported to AvaTax (after)  │ 312.64       │ 27.36 │ 340.00 │
  └─────────────────────────────┴──────────────┴───────┴────────┘
</pre>

The derived rate is also corrupted: the module stores taxCalculated / taxableAmount, so the example above yields a stored rate of 8.7529% instead of the real 8.75%.

### **Fix**

Report price_subtotal, the net line total Odoo already computes, instead of recomputing the base from price_unit:

- account.move.line._compute_totals sets price_subtotal = taxes_res['total_excluded']
- sale.order.line._compute_amount sets price_subtotal = amount_untaxed from _compute_taxes()

Reading that field instead of recomputing has two benefits: the reported amount is by construction the same figure the document shows, so it cannot drift from it, and it does not re-enter account.tax.compute_all, which this module patches.

_get_avatax_amount keeps its signature, its sign handling and its currency conversion. The documented qty argument still works and is now also net of included  taxes.

Covered by 18 tests across both modules, including backward compatibility against the previous formula, mixed included and excluded taxes on one line, credit note sign, and a round trip asserting the base converges when AvaTax writes its tax back.

### Why both modules in a single PR

The invoice copies price and taxes from the order line without recomputing them, so the two paths are not independently correct. Splitting the fix leaves the order to invoice  channel broken, which is measurable: with only the account_avatax_sale_oca change applied, the order line reports the net base while the invoice generated from it still reports the gross amount.

  ```
  FAIL: TestSaleOrderBaseAmount.test_05_sale_amount_matches_invoice
  AssertionError: 625.29 != 680.0 within 2 places
  ```

 I can split it into two PRs if preferred, at the cost of that cross-path test.

### **Alternative considered: the taxIncluded flag**

AvaTax LineItemModel does expose taxIncluded ("Indicates whether the amount for this line already includes tax"), and sending the gross amount together with that flag would be a  valid design. It was rejected because it requires deciding, per line, whether the amount is gross, and that decision is not reliable:

- With override_line_taxes enabled, AvaTax replaces the line tax with one of its own, copied from the 0% AvaTax template. If that template is not price included, an any(price_include) predicate reads False on a line whose price_unit is still gross.
- A single boolean cannot describe a line that mixes an included and an excluded tax.

### Note on the third commit
  
 The last commit fixes a pre-existing test in this module rather than the defect above. It is  included because CI cannot pass without it: TestAvatax.test_avatax_compute_tax creates an  avalara.salestax record with placeholder credentials, and service_url defaults to the  production endpoint. get_avatax_config_company and create_transaction are mocked, but  void_transaction is not, so button_draft() reaches the live service and fails  authentication.

This is not a regression from the change above. It reproduces on a clean 16.0 checkout with  none of the files from this PR present, and it fails identically — same traceback, same line  numbers. The green runs of that test are all months old (the most recent is 2026-03-02), while  this branch failed twice on 2026-07-29 with byte-identical trees, so the live service appears to  answer differently now than it used to.

The fix disables document recording, which is what the flag is for in a test setup, so no  request is attempted at all. Because disabling it also makes the document type provisional, that  change of scenario is asserted explicitly rather than left as an invisible side effect.

### DIFF
--- a/account_avatax_oca/models/account_move.py
+++ b/account_avatax_oca/models/account_move.py
@@ -461,32 +461,45 @@ class AccountMoveLine(models.Model):
         Return the company currency line amount, after discounts,
         to use for Tax calculation.
 
-        Can be used to compute unit price only, using qty=1.
+        Taxes flagged as price included are embedded in ``price_unit``, so the
+        raw line amount is gross and must not be reported as a taxable base.
+        ``price_subtotal`` is the net line total Odoo already computes, and it
+        equals ``price_unit * quantity * (1 - discount / 100)`` when none of the
+        line taxes is price included.
 
-        Code extracted from account/models/account_move.py,
-        from the compute_base_line_taxes() nested function,
-        adjusted to compute line amount instead of unit price.
+        Can be used to compute unit price only, using qty=1.
         """
         self.ensure_one()
         base_line = self
         move = base_line.move_id
         sign = -1 if move.is_inbound() else 1
-        quantity = qty or base_line.quantity
-        base_amount = base_line.price_unit * quantity
-        if base_line.currency_id:
-            price_unit_foreign_curr = (
-                sign * base_amount * (1 - (base_line.discount / 100.0))
+        if not qty:
+            base_amount = base_line.price_subtotal
+        else:
+            discounted_price_unit = base_line.price_unit * (
+                1 - (base_line.discount / 100.0)
             )
+            if base_line.tax_ids:
+                base_amount = base_line.tax_ids.compute_all(
+                    discounted_price_unit,
+                    currency=base_line.currency_id,
+                    quantity=qty,
+                    product=base_line.product_id,
+                    partner=base_line.partner_id,
+                    is_refund=base_line.is_refund,
+                )["total_excluded"]
+            else:
+                base_amount = discounted_price_unit * qty
+        base_amount = sign * base_amount
+        if base_line.currency_id:
             price_unit_comp_curr = base_line.currency_id._convert(
-                price_unit_foreign_curr,
+                base_amount,
                 move.company_id.currency_id,
                 move.company_id,
                 move.date,
             )
         else:
-            price_unit_comp_curr = (
-                sign * base_amount * (1 - (base_line.discount / 100.0))
-            )
+            price_unit_comp_curr = base_amount
         return -price_unit_comp_curr
 
     # Same in v12
--- a/account_avatax_sale_oca/models/sale_order.py
+++ b/account_avatax_sale_oca/models/sale_order.py
@@ -334,9 +334,12 @@ class SaleOrderLine(models.Model):
         else:
             item_code = product.default_code or ("ID:%d" % product.id)
         tax_code = line.product_id.applicable_tax_code_id.name
-        amount = (
-            sign * line.price_unit * line.product_uom_qty * (1 - line.discount / 100.0)
-        )
+        # Taxes flagged as price included are embedded in price_unit, so the raw
+        # line amount is gross and must not be reported as a taxable base.
+        # price_subtotal is the net line total Odoo already computes, and it
+        # equals price_unit * product_uom_qty * (1 - discount / 100) when none of
+        # the line taxes is price included.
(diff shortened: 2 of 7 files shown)


aritzolea (approved):

dreispt:
Is the point for this to support "tax included" prices? That's not clear from the PR title.
In the US prices are always tax excluded, or is there a particular case where it is not so?
So, I'm looking for the rationale behind these proposed changes.

account_avatax_oca/models/account_move.py
```diff
@@ -456,37 +456,51 @@ class AccountMoveLine(models.Model):
 
     avatax_amt_line = fields.Float(string="AvaTax Line", copy=False)
 
+    def _get_avatax_base_amount(self, qty=None):
```
dreispt: This is really the same as `_get_avatax_amount`, I don't think a new method is needed.

account_avatax_sale_oca/models/sale_order.py
```diff
@@ -319,6 +319,18 @@ class SaleOrderLine(models.Model):
...
+        """
+        Return the net (tax excluded) line total.
+
+        Taxes flagged as price included are embedded in ``price_unit``, so the
+        raw line amount is gross and must not be reported as a taxable base.
+
+        A line total, not a unit price.
+        """
+        self.ensure_one()
+        return self.price_subtotal
```
dreispt: Methods is too simple, looks redundant.

almumu:
> Is the point for this to support "tax included" prices? That's not clear from the PR title. In the US prices are always tax excluded, or is there a particular case where it is not so? So, I'm looking for the rationale behind these proposed changes.

@dreispt  Right, this is not how US sales tax normally works. The setup is a global shop whose standard is tax inclusive pricing, integrated with Odoo. The order line ends up with a gross `price_unit` and a price included tax on it, while AvaTax is what determines the tax.

The module already handles price included taxes: `post_load_hook` replaces `_compute_all_tax` and passes `handle_price_include=True`, so the Odoo document  totals are net. Only the reported amount does not follow, because it recomputes from `price_unit`, which is gross, so the document and the AvaTax payload  disagree with each other. On a 340 gross line at 8.75% included, Odoo shows 312.64 + 27.36 while AvaTax gets a base of 340. It also stores the derived rate as 8.7529% instead of 8.75%, which is how it surfaced.

`price_subtotal` equals the previous formula when no line tax is price included, so nothing changes for existing installations  there is a test pinning  that.

I have retitlled the pr and simplified the redundant methods.

PR #609, opened 2026-08-26
[18.0] account_tax, account_ecotax_sale: add ecotax only if it matches shipping address country

rvalyi:
Forward-port to 18.0 of the ecotax-by-country feature merged in 16.0 by
Florian da Costa (OCA#501), following the unmerged 17.0 port attempt
OCA#514. Split into 2 commits, one per module:

- account_ecotax: `country_ids` on account.ecotax.classification (empty =
  all countries), `_get_country_eligible_classification()` on
  product.product, ecotax lines filtered by the delivery
  (partner_shipping/partner) country on invoice lines, helper
  `_get_ecotax_amounts_from_classification()` extracted. Includes the
  c482033d reliability fix (`self and self.ensure_one()`) from the 16.0
  merged state. New test_06_ecotax_by_country.
- account_ecotax_sale: same country filtering on sale order lines,
  new test_03_ecotax_by_country.

Adaptations to 18.0:
- weight-based ecotax amounts are still recomputed from the product weight
  (18.0 behavior, kept inside the extracted helper)
- the sale test triggers `_compute_ecotax_line_ids()` explicitly after Form
  save, following the existing test_01/test_02 pattern (computed O2m is not
  persisted through Form in 18.0)

The account_ecotax_tax / account_ecotax_sale_tax country support (also part
of OCA#501 in 16.0, with a different computation approach on 18.0) will be
proposed in a follow-up PR.

### DIFF
--- a/account_ecotax/models/account_ecotax_classification.py
+++ b/account_ecotax/models/account_ecotax_classification.py
@@ -82,6 +82,12 @@ class AccountEcotaxClassification(models.Model):
     )
     intrastat_code = fields.Char()
     scale_code = fields.Char()
+    country_ids = fields.Many2many(
+        "res.country",
+        string="Countries",
+        help="Ecotax will be applied when delivered in the listed countries (or all "
+        "countries if empty)",
+    )
 
     @api.depends("ecotax_type")
     def _compute_ecotax_vals(self):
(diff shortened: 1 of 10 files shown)


hparfr (approved):
LGTM

